PR #16, opened 2015-01-28
Install gcc in Dockerfile

discordianfish:
With procfs we now depend on gcc.
This also removes vim-common which isn't used.


juliusv:
:+1:

grobie:
If anyone knows how to get the clock ticks in go without importing the "C" package ...

brian-brazil:
> If anyone knows how to get the clock ticks in go without importing the "C" package ...

That was the only way I found when looking that up for the node_exporter.

Issue #18, opened 2015-01-30
Consider not assigning an instance label if none is requested


beorn7:
Currently, we add the ip number:
https://github.com/prometheus/pushgateway/blob/master/handler/push.go#L55

Cf. https://github.com/prometheus/prometheus/issues/490

beorn7:
Currently, everything in the pushgateway is grouped by job/instance tuples... So not having an instance label needs some thinking. Probably treat an empty instance label as 'no label' (and do not export it in that case) so that all metrics with only a job label are in the same group then.

brian-brazil:
A thought I had: if we do this we could change it so `instance` isn't special and /job/myjob/label/myLabel/ would set `label`. That could also act as a grouping for UI purposes, though I'm not too worried about the grouping/UI overall.

beorn7:
I guess, concluding from other discussions, we should get completely rid of the special role of the instance label and never meddle with the instance label, remove it from the REST API, and don't group by it.

beorn7:
Migration path: Encourage users to not set an instance label. Stop setting one automatically, and group by job only whatever has no instance label. (An instance label set as a 'real' labes, i.e. not via the REST API, could be ignored for grouping... Currently it is _overwritten_...)
Once nobody sets instance labels anymore, we can just remove them completely from the pushgateway and from clients (except as normal labels, of course).

grobie:
Will it be possible for clients to set the instance label in a different way, for example like any other label? What is with use cases where a process can't be scraped (workers where it's unfeasible to start a HTTP server, services behind strict firewalls, etc.), but where an instance label is still desired?

beorn7:
Yes, an instance label would be a label as any other. (With the only distinction that the Prometheus server sets it... but we already discussed to make that configurable, see https://github.com/prometheus/prometheus/issues/490 .)

The important thing is that the pushgateway will not set one on its own, and there will be no grouping by instance labels. DELETE will delete all metrics of a job, and PUT will replace all metrics of a job.

grobie:
Without the option to define a grouping label, it will be very expensive (many HTTP calls), complicated (the client needs to check any previously set label to override it/delete it) and racy (it's possible that the server scrapes the pgw while a client is still doing POST/DELETEs) for multiple instances to push metrics to the pgw under the same job label. Essentially, it means the pushgateway is not longer suitable to act as gateway for multiple instances which can't be scraped and need to push their metrics instead.

[...]

PR #22, opened 2015-02-08
Auto-fill instance with IPv6 address (instead of incorrect "[").

aecolley:
Sending to a pushgateway on localhost, without specifying an instance field, caused pushgateway to fill in `instance="["`. This is because the IP address is `[::1]`, whereas the code assumes an IPv4 address. Using the standard `net.SplitHostPort` is the right thing, and it produces the right results (namely, `instance="::1"`).

### DIFF
--- a/handler/push.go
+++ b/handler/push.go
@@ -16,8 +16,8 @@ package handler
 import (
 	"io"
 	"mime"
+	"net"
 	"net/http"
-	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


juliusv:
Thanks, :+1:! Just FYI, we might completely omit instance labels in the future if they're not explicitly set by the client (https://github.com/prometheus/pushgateway/issues/18).

PR #30, opened 2015-03-15
Remove internal metrics.

beorn7:
With current versions of client_golang, internal metrics are
automatically exported by the client library. (In fact, not all mem
stats are included yet, but that will be added eventually. See
https://github.com/prometheus/client_golang/issues/96 .)

@juliusv


juliusv:
:+1:

Issue #37, opened 2015-05-05
Makefile requires go-bindata but doesn't install it


discordianfish:
Related to #29

beorn7:
Actually, it does... Can you tell me how to reproduce the problem?

discordianfish:
@beorn7 TBH, no idea how I ran into it. I'll close this and reopen if I run into again. Didn't provide more details because I though we simply forgot to install it.

beorn7:
I think @jcftang reproduced it. I guess if the go tool in .deps is used, go-bindata gets installed there and not into GOPATH, or something. Has to be investigated. Output:

jtang@awesome-o-w:[(none)@master]pushgateway$ make
GOROOT=/Users/jtang/develop/pushgateway/.deps/go GOPATH=/Users/jtang/develop /Users/jtang/develop/pushgateway/.deps/go/bin/go get -d
GOROOT=/Users/jtang/develop/pushgateway/.deps/go GOPATH=/Users/jtang/develop /Users/jtang/develop/pushgateway/.deps/go/bin/go get github.com/jteeuwen/go-bindata/...
/Users/jtang/develop/bin/go-bindata resources/
make: /Users/jtang/develop/bin/go-bindata: No such file or directory
make: *** [bindata.go] Error 1
jtang@awesome-o-w:[(none)@master]pushgateway$ git describe
fatal: No annotated tags can describe 'eefacc22c6a9b6393dde154804d65094cd54b3ed'.
However, there were unannotated tags: try --tags.

jcftang:
I had some issues with installing go-bindata on OSX, I've yet to try on a debian/ubuntu box.

beorn7:
https://github.com/prometheus/pushgateway/pull/40 should really solve this...

PR #38, opened 2015-05-27
Never honor GOBIN and let the GOBIN path set automatically

jcftang:
This attempts to fix https://github.com/prometheus/pushgateway/issues/37

In my environment I had $GOBIN set, this confused the makefile into installing the go-bindata binary into the wrong path. Unsetting the $GOBIN variable lets the build complete in a more reliable way.


beorn7:
Sure, I can merge this. But I'm currently working on switching the whole pushgateway code to using https://github.com/prometheus/utils/blob/master/Makefile.COMMON , which will implicitly solve this problem.

Merging anyway, just that you know...

jcftang:
Ok, thanks!

Issue #43, opened 2015-06-08
Cross compilation fails due to go-bindata not in GOPATH


discordianfish:
When cross-compiling, the Makefile installs go-bindata in `.build/gopath/bin/linux_386/go-bindata` but tries to call it via `.build/gopath/bin/go-bindata`:

```
$ GOOS=linux GOARCH=386 make archive
GOPATH=/home/fish/dev/go/src/github.com/prometheus/pushgateway/.build/gopath /home/fish/dev/go-src/bin/go get github.com/jteeuwen/go-bindata/...
/home/fish/dev/go/src/github.com/prometheus/pushgateway/.build/gopath/bin/go-bindata -prefix=resources resources/...
make: /home/fish/dev/go/src/github.com/prometheus/pushgateway/.build/gopath/bin/go-bindata: Command not found
Makefile:33: recipe for target 'bindata.go' failed
make: *** [bindata.go] Error 127
```

sdurrheimer:
Fixed by #76

PR #45, opened 2015-06-24
Release 0.2.0

discordianfish:
Due to the major change to support arbitrary grouping, this deserves a
MINOR version increment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.2.0 / 2015-06-25
+* [CHANGE] Support arbitrary grouping of metrics.
+
 ## 0.1.2 / 2015-06-08
 * [CHANGE] Move pushgateway binary in archive from bin/ to /.
 * [CHANGE] Migrate logging to prometheus/log.
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-VERSION  := 0.1.2
+VERSION  := 0.2.0
 TARGET   := pushgateway
 
 REV        := $(shell git rev-parse --short HEAD)


brian-brazil:
:+1:

beorn7:
There is actually a breaking change in the DELETE API call.

The version increment is fine, but the breaking change needs to be documented.

beorn7:
I have changed the changelog, and will set the tag to after that change.

beorn7:
@discordianfish I have updated the release tag and info. Could you use your magic script to attach binaries? (Apparently, it doesn't work on Linux...)

discordianfish:
@beorn7 The magic script isn't working due to #43 :-/
And sorry for not realizing this is a breaking change. Agreed that it should be included in the changelog. Maybe come up with some general guidelines how to manage those things. I think it would make sense for example to update the changelog in the PRs introducing the change. That way it wouldn't be possible to forget mentioning that in a release.

juliusv:
In my experience it's much more reliable to collect changelog entries once when you cut the release than to try and think of updating the changelog with every PR (and also remember to make all external contributors do that). Further, the changelog often doesn't end up with per-PR updates, but a lot of stuff is summarized/squashed together in the end.

discordianfish:
@juliusv Hrm.. but I think at least for breaking change it would be a good idea.

grobie:
My experience is the opposite of @juliusv.

beorn7:
Ideally, the person who knows the repository in question best should be involved when cutting the release and/or creating the changelog.

discordianfish:
Whatever we decide, we should document it somewhere. I prefer adding things to the changelog with the PR, it's also useful for people building from master to have a up-to-date changelog. But I leave this to you.

Issue #46, opened 2015-06-28
Dockerfile broken


Luzifer:
- [Build-Status on DockerHub](https://registry.hub.docker.com/u/prom/pushgateway/builds_history/129787/)
- Build result

```
[...]
Running hooks in /etc/ca-certificates/update.d....done.
 ---> 9c2a82406edb
Removing intermediate container 926c3b659c6f
Step 6 : ADD . /pushgateway
 ---> c061737f1fe8
Removing intermediate container 325ba4628026
Step 7 : RUN make bin/pushgateway
 ---> Running in 35d739a885ff
make: *** No rule to make target `bin/pushgateway'.  Stop.
INFO[0088] The command [/bin/sh -c make bin/pushgateway] returned a non-zero code: 2
```

juliusv:
Thanks! Fix PR is out in https://github.com/prometheus/pushgateway/pull/47 to update the Dockerfile to use the new Makefile.

PR #47, opened 2015-06-29
Fix Dockerfile make and entrypoint steps.

juliusv:
Fixes https://github.com/prometheus/pushgateway/issues/46


juliusv:
@discordianfish

jayunit100:
shall we merge this @brian-brazil  @discordianfish ?  looks like it was broken in tag 0.2.0 also i think

discordianfish:
LGTM but we should replaced this by @sdurrheimer's base image soon.

PR #49, opened 2015-07-09
Add example for building docker and starting  

jayunit100:
just a minor example for newcomers that should copy and paste work ootb


beorn7:
Thanks for the PR.

I think this is superseeded  by https://github.com/prometheus/pushgateway/pull/50/ .

Please re-open if you disagree.

PR #52, opened 2015-08-05
Remove required labels flag.

brian-brazil:
When initally putting in this logic we were unsure if
there were labels other than instance that could need to be
set to empty. After further thought, it's only the job
and instance labels that are mandatory on the prometheus server
side and as job is already mandatory for the pushgateway
only instance needs special casing.

@beorn7

### DIFF
--- a/handler/push.go
+++ b/handler/push.go
@@ -27,6 +27,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/julienschmidt/httprouter"
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
+	"github.com/prometheus/client_golang/model"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/text"
 
@@ -41,7 +42,7 @@ import (
 //
 // The returned handler is already instrumented for Prometheus.
 func Push(
-	ms storage.MetricStore, replace bool, requiredLabels map[string]struct{},
+	ms storage.MetricStore, replace bool,
 ) func(http.ResponseWriter, *http.Request, httprouter.Params) {
 	var ps httprouter.Params
 	var mtx sync.Mutex // Protects ps.
@@ -98,11 +99,7 @@ func Push(
 				http.Error(w, err.Error(), http.StatusInternalServerError)
 				return
 			}
-			sanitizeLabels(
-				metricFamilies,
-				labels,
-				requiredLabels,
-			)
+			sanitizeLabels(metricFamilies, labels)
 			ms.SubmitWriteRequest(storage.WriteRequest{
 				Labels:         labels,
 				Timestamp:      time.Now(),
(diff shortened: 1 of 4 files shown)


beorn7:
It was obvious that currently only job and instance labels are mandatory on the prometheus server. The flag was only ever meant for special cases or future changes.
What has changed your mind now?

brian-brazil:
I've thought about it and realised that there was in reality no need to make this configurable.

handler/push.go
```diff
@@ -207,20 +200,20 @@ func LegacyPush(
...
 //
 // Finally, sanitizeLabels sorts the label pairs of all metrics.
 func sanitizeLabels(
 	metricFamilies map[string]*dto.MetricFamily,
-	groupingLabels map[string]string, requiredLabels map[string]struct{},
+	groupingLabels map[string]string,
 ) {
 	gLabelsNotYetDone := make(map[string]string, len(groupingLabels))
-	rLabelsNotYetDone := make(map[string]struct{}, len(requiredLabels))
+	rLabelsNotYetDone := make(map[string]struct{}, 1)
```
beorn7: We don't really need a map here anymore. It's only the check for instance now, so the logic below should be simplified accordingly.

brian-brazil: Changed, though it's more code.

handler/push.go
```diff
@@ -207,20 +200,19 @@ func LegacyPush(
...
 }
 
-// sanitizeLabels ensures that all the labels in groupingLabels and
-// requiredLabels are present in each MetricFamily in metricFamilies. The label
-// values from groupingLabels are set in each MetricFamily, no matter what. In
-// contrast, requiredLabels contains only label names and no label values. Only
-// if any of those labels are not present at all in a MetricFamily, they will be
-// created (with an empty string as value).
+// sanitizeLabels ensures that all the labels in groupingLabels and the
+// `instance` label is present in each MetricFamily in metricFamilies. The
```
beorn7: Nit: is -> are.

handler/push.go
```diff
@@ -207,20 +200,19 @@ func LegacyPush(
...
+// label values from groupingLabels are set in each MetricFamily, no matter
+// what. In contrast, 'instance' will have no label values. Only if any of
+// those labels are not present at all in a MetricFamily, they will be created
+// (with an empty string as value).
```
beorn7: This needs more rewording. Instead of everything including and after "In contrast", I suggest:

After that, if the 'instance' label is not present at all in a MetricFamily, it will be created (with an empty string as value).

beorn7:
Besides comment nits, :+1:

handler/push.go
brian-brazil: That's better alright, updated.

handler/push.go
brian-brazil: Done

PR #53, opened 2015-08-20
Set REV and BRANCH to unknown if cwd is not a git repo

xbglowx:
The build fails when the source directory is not managed
by git, i.e. release download.

@beorn7

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,8 @@
 VERSION  := 0.2.0
 TARGET   := pushgateway
 
-REV        := $(shell git rev-parse --short HEAD)
-BRANCH     := $(shell git rev-parse --abbrev-ref HEAD)
+REV        := $(shell git rev-parse --short HEAD 2> /dev/null  || echo 'unknown')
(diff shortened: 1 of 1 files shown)


fabxc:
Those two lines include the branch and revision information into the final build. This is something we want as it allows users to tell us the exact code their binary is running when reporting problems.
It's also something we use in all Prometheus components.

Why is it causing a problem for you, i.e. why do you not simply keep the git folder?
Just removing stuff is not a valid solution, regardless.

juliusv:
For `prometheus/prometheus`, this was solved like this: https://github.com/prometheus/prometheus/pull/806

Maybe solve it similarly here?

xbglowx:
@fabxc I didn't remove the git folder. That is the way it came if you download the source using the download link.

@juliusv That looks like a good solution. Let me update my commit.

fabxc:
@xbglowx yes, I always forget the GitHub zips repos for you. Good that there's another solution.

xbglowx:
@fabxc @juliusv commit updated.

fabxc:
Thanks! :+1:

xbglowx:
Thanks!

PR #58, opened 2015-09-18
Fix direct acess of protobuf fields

beorn7:
This also updates the Makefile. Sorry for mixing things up.

@fabxc @juliusv


juliusv:
:+1:

PR #59, opened 2015-09-27
Reduce flakiness in time-based tests.

brian-brazil:
With the new GC, 10ms is the minimum we should be sleeping
to wait for simple operations, so use 20ms to be safe.
The failure I saw was in TestNoPersistence.

@beorn7


beorn7:
:+1: (with the usual caveats that "proper" tests should synchronize in a reproducible way).

PR #62, opened 2015-10-16
Update port number in readme

### LINKED ISSUE #61: Basic setup of the pushgateway
Hi, I built the pushgateway from git and started the binary as described. From my understanding one uses port 8080 for pushing the metrics using http and port 9091 to let prometheus pull them from the gateway. In my case the curl examples are not working. Netstat shows me no sign of usage for port 8080. I guess i do something fundamentally wrong here. It would be nice if someone can help me out :/.

brian-brazil:
@beorn7 

Fixes #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,13 +52,13 @@ Examples:
 
 * Push a single sample into the group identified by `{job="some_job"}`:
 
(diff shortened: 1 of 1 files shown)


juliusv:
:+1:

PR #63, opened 2015-10-27
Fix lint warnings

akfaew:
Fix these:
link: warning: option -X main.buildVersion 0.2.0 may not work in future releases; use -X main.buildVersion=0.2.0
link: warning: option -X main.buildRev 9647eef may not work in future releases; use -X main.buildRev=9647eef
...


fabxc:
Thanks :+1:!

PR #64, opened 2015-11-04
Push should reject requests with improper labels.

jpittis:
prometheus/pushgateway/issues/54 explains a bug where a labelname of `a=b` was not rejected by pushgateway.

The current iteration of this patch checks all labelnames against the official regex provided in the documentation: `[a-zA-Z_][a-zA-Z0-9_]*`. It comes with a test to confirm `splitLabels` is functioning correctly.

@grobie mentioned in irc that returning a `4xx` would be a good solution. He also mentioned that it would be good to increment some kind of error metric when this error was encountered. I'm not sure what he means by the latter statement.

This would be my first attempt at a patch to Prometheus. Please state frankly if this is more of a burden then of help.

### DIFF
--- a/handler/push.go
+++ b/handler/push.go
@@ -265,7 +265,12 @@ func splitLabels(labels string) (map[string]string, error) {
 	if len(components)%2 != 0 {
 		return nil, fmt.Errorf("odd number of components in label string %q", labels)
 	}
+
 	for i := 0; i < len(components)-1; i += 2 {
+		if !model.LabelNameRE.MatchString(components[i]) ||
+			strings.HasPrefix(components[i], model.ReservedLabelPrefix) {
+			return nil, fmt.Errorf("improper label name %q", components[i])
+		}
 		result[components[i]] = components[i+1]
 	}
 	return result, nil
--- a/handler/handler_test.go
+++ b/handler/handler_test.go
@@ -401,3 +401,36 @@ func TestDelete(t *testing.T) {
 		t.Errorf("Wanted instance %v, got %v.", expected, got)
 	}
 }
+
+func TestSplitLabels(t *testing.T) {
+	proper_labels := "/label_name1/label_value1/label_name2/label_value2"
+	expected_parsed := map[string]string{
+		"label_name1": "label_value1",
+		"label_name2": "label_value2",
+	}
+	parsed, err := splitLabels(proper_labels)
+	if err != nil {
+		t.Errorf("Got unexpected error: %s.", err)
+	}
+	for k, v := range expected_parsed {
+		got, ok := parsed[k]
+		if !ok {
+			t.Errorf("Expected to find key %s.", k)
+		}
+		if got != v {
+			t.Errorf("Expected %s but got %s.", v, got)
+		}
+	}
+
+	improper_labels := "/label_name1/label_value1/a=b/label_value2"
+	_, err = splitLabels(improper_labels)
+	if err == nil {
+		t.Error("Expected splitLabels to return an error when given improper labels.")
+	}
+
+	reserved_labels := "/label_name1/label_value1/__label_name2/label_value2"
+	_, err = splitLabels(reserved_labels)
+	if err == nil {
+		t.Error("Expected splitLabels to return an error when given a reserved label.")
+	}
+}


handler/push.go
```diff
@@ -36,6 +37,9 @@ import (
 	"github.com/prometheus/pushgateway/storage"
 )
 
+// Regex labelRegex is used to check the validity of labelnames in splitLabels.
+var labelRegex = regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]*$")
```
brian-brazil: You should use `LabelNameRE` from https://github.com/prometheus/common/blob/master/model/labels.go#L84 rather than duplicating this

brian-brazil:
Change generally looks good, thanks!

grobie:
My suggestion was to register a prometheus metric like `pushgateway_push_invalid_total` which would be a counter and be incremented every time we detect an invalid push request. That would allow users to create an alert if some systems repeatedly tries to push something invalid.

jpittis:
@grobie: Seems like an important metric to have. Would this best be done in a separate PR as push failures can occur for multiple reasons? (Besides the invalidation added in this PR, there are others like the one [here](https://github.com/jpittis/pushgateway/blob/06f8ba42a06512f0074ab86b3c17e1047c911c7a/handler/push.go#L266) and [here](https://github.com/jpittis/pushgateway/blob/06f8ba42a06512f0074ab86b3c17e1047c911c7a/handler/push.go#L63) that would also need to increment the metric.)

brian-brazil:
It's probably best to leave that to another PR.

grobie:
As I couldn't immediately see any instrumentation in the pgw, that sounds
good to me.

jpittis:
Should the pushgateway also reject labels that start with a `__`?

grobie:
@jpittis I just realized that the pushgateway uses the [prometheus.InstrumentedHandler](https://github.com/prometheus/pushgateway/blob/master/handler/push.go#L50). So the pushgateway already exports a `http_requests_total` with a `code` label. That should be enough for users to set up an alert on 400 responses.

brian-brazil:
> Should the pushgateway also reject labels that start with a __ as they are meant for internal use?

We reject them in the clients, so it'd make sense to reject here too.

beorn7:
Instrumentation of the Pushgateway surely needs some thorough review. But I first want to fix `client_golang` including the InstrumentedHandler convenience wrapper and then check how to best handle it here.

handler/push.go
```diff
@@ -265,7 +265,12 @@ func splitLabels(labels string) (map[string]string, error) {
 	if len(components)%2 != 0 {
 		return nil, fmt.Errorf("odd number of components in label string %q", labels)
 	}
+
 	for i := 0; i < len(components)-1; i += 2 {
+		if !model.LabelNameRE.MatchString(components[i]) ||
+			strings.HasPrefix(components[i], model.ReservedLabelPrefix) {
+			return nil, fmt.Errorf("improper label name %s", components[i])
```
beorn7: Nitpick: Better to use `%q` here (as [a few lines above](https://github.com/prometheus/pushgateway/pull/64/files#diff-afffdd5e76af154b0c3d69694a63a918R266)).

beorn7:
:+1: besides my tiny nit.

jpittis:
8a20460 adds fixes @beorn7's suggestion. It also squashes the commits into one.

(On a side note, when making changes based on a review from a PR, is it better to commit every little change and only squash them upon merge? Or is it alright if the PR owner force pushes their branch? I guess I'm wondeirng if a change like @beorn7's "tiny nitpick" should have been done as a seperate commit rather than squashing it into the previous commits.)

brian-brazil:
In Prometheus we generally ask for changes to be squashed just before we merge them.

beorn7:
Opinions vary a bit among the Prometheus developers…

I definitely prefer to see the individual commits during the review to track how the code review has developed.

I tend to dislike squashing. (Every time you rebase or force-push, Linus Torvalds kills a kitten…) If the number of commits are excessive, or the review has only taken a short time, I can be convinced to squash just before merge. However, especially if a review drags on for a long time, other people might already use commits from the review branch. Squashing will make their life really hard, as it breaks the fundamental idea of a distributed VCS.

juliusv:
I'm totally for squashing before merge unless every commit really is a major thing on its own and/or was meant to be a separate commit to begin with. This is also common practice.

If you think about how Git is used in the Linux kernel, fixups that happen *during* a code review also don't end up as separate commits in the final history (though there the initial submission of changes happens via patches in emails, not via PRs, the resulting commit structure is the same as squashing here). You could argue that GitHub does code reviews wrong in general by basing them on branches rather than clearly defined and versioned changesets (like Gerrit does).

grobie:
I'm totally in favor of squash commits into reasonable self-explaining commits (doesn't mean it only has to be a single commit). Reading the commit history is a lot easier if each commit includes tests/documentation changes and all nitpicks changes then to have a lot of small one line change commits in master.

beorn7:
I understand all the arguments for squashing, and I don't doubt their value.

But there is a trade-off.

The problem is that code reviews happen in the public, so commits that happen during a code review could in principle be used by others. It's open source, and it's a distributed VCS. The golden rule is: Once a commit has been pushed "out there", you must not rebase anymore. In principle, once a commit is out there on a public repo in github, it's "out there". In practice, it's most of the time not a problem, because commits are not immediately used by others. But the conceptual dissonance between the various concepts should be kept in mind.

Sometimes it matters in practice. Let's say I write my own cool now Prometheus feature. I share it with others for testing. And finally I submit a PR to `prometheus/prometheus`. Once my changes are in, I can tell all my co-workers that helped me develop the feature, they can finally merge with `prometheus/prometheus` again. They will get very upset if all the commits have changed and they have to do CVS style merges as if it were still 1995. So in that case, I'd insist to have the code-review changes on top of my existing commits. However, due to the thoughtless "please squash/rebase" thing, that usually requires lengthy arguing, explaining from the beginning what a DVCS is all about…

jpittis:
Thanks for the advice everyone. :)

Issue #69, opened 2016-01-13
cloning + making + running = not working


jebl01:
> Note! complete golang noob speaking...

Perhaps there is something more to it, but in that case I think that the readme needs some work :-)

I can make and run, but the result is not pretty and the logs are being filled with logs like the one below:
```
2016/01/13 16:07:53 http: panic serving [::1]:51642: runtime error: invalid memory address or nil pointer dereference
goroutine 102 [running]:
net/http.(*conn).serve.func1(0xc82049e2c0, 0xe00000, 0xc82049c020)
	/usr/local/go/src/net/http/server.go:1287 +0xb5
github.com/elazarl/go-bindata-assetfs.(*AssetFS).Open(0xc8203b7500, 0xc820450035, 0x13, 0x0, 0x0, 0x0, 0x0)
	/Users/jesblo/dev/gopath/src/github.com/prometheus/pushgateway/.build/gopath/src/github.com/elazarl/go-bindata-assetfs/assetfs.go:148 +0x22e
net/http.serveFile(0xba4758, 0xc820030000, 0xc8204d4000, 0xb64ce0, 0xc8203b7500, 0xc820450034, 0x14, 0x1)
	/usr/local/go/src/net/http/fs.go:359 +0x158
net/http.(*fileHandler).ServeHTTP(0xc82034b3b0, 0xba4758, 0xc820030000, 0xc8204d4000)
	/usr/local/go/src/net/http/fs.go:483 +0x19c
net/http.(Handler).ServeHTTP-fm(0xba4758, 0xc820030000, 0xc8204d4000)
	/Users/jesblo/dev/gopath/src/github.com/prometheus/pushgateway/.build/gopath/src/github.com/prometheus/client_golang/prometheus/http.go:61 +0x50
github.com/prometheus/client_golang/prometheus.InstrumentHandlerFuncWithOpts.func1(0xba4680, 0xc8201f2000, 0xc8204d4000)
	/Users/jesblo/dev/gopath/src/github.com/prometheus/pushgateway/.build/gopath/src/github.com/prometheus/client_golang/prometheus/http.go:158 +0x335
net/http.HandlerFunc.ServeHTTP(0xc820273e00, 0xba4680, 0xc8201f2000, 0xc8204d4000)
	/usr/local/go/src/net/http/server.go:1422 +0x3a
[...]

Issue #29, opened 2015-02-25
Vendor all depencies and refactor Makefile...


beorn7:
... so it matches the way it is done in other Prometheus projects.

grobie:
Could have prevented #35.

beorn7:
https://github.com/prometheus/pushgateway/pull/40 switches to `Makefile.COMMON`. I'll postpone the vendoring until we have clarity which direction to go (godep? gb? something completely different?).

grobie:
@beorn7 Any conclusion here already? GOVENDOREXPERIMENT (+ godep)?

beorn7:
Same as in prometheus/prometheus: GOVENDOREXPERIMENT, and use the dependency management tool of your choice as long as it creates what is compatible with GOVENDOREXPERIMENT.

beorn7:
Fixed thanks to @sdurrheimer

PR #70, opened 2016-01-14
Go update and adjusting to upstream changes

beorn7:
@jebl01 @grobie (because you are still at work ;)

Fixes https://github.com/prometheus/pushgateway/issues/69


grobie:
:+1: (first wanted to complain about the missing vendoring, but #29 is there already :) ).

beorn7:
And https://github.com/prometheus/pushgateway/issues/29  is also mentioned in https://github.com/prometheus/pushgateway/issues/69 . :)

We are ready to fix https://github.com/prometheus/pushgateway/issues/29, BTW. But other priorities right now on my side…

PR #92, opened 2016-09-05
Improve README.md

### LINKED ISSUE #87: Document the "distributed counter" non-goal
Quite often, people think of the Pushgateway as a distributed counter. Then they realize it's not doing that, and file a feature request.

We should make clear in the README.md that the PGW is not a distributed counter, and that the functionality might be something implemented as a stand-alone tool or (much less likely) as an additional feature in the PGW.

### LINKED ISSUE #89: Add documentation for adding release events
In various graphing tools (such as Grafana), it is useful to tag graphs with events, such as a release deployment. This way you can quickly see why a particular metric has suddenly jumped up  / down.

The prometheus push gateway is an ideal tool for pushing a single event on a successful deployment, as the deployment is typically triggered from a ephemeral single-run task on a build agent. 

With the available metrics in prometheus, it is unclear what the best method of storing this data is, and how exactly to push into the push gateway in such a way to allow annotation of events on metric graphs.

Right now, we're trying to push a metric as a counter, but this is appearing every minute in our graphing, presumably due to the push gateway being scraped every minute.

It would be nice if this (fairly common) use case was covered, even briefly, in the documentation. (even if the answer was: "We don't have great support for this yet, but you can use xyz in the meantime")

beorn7:
This addresses three issues:

- Mention binary releases.
- Make even clearer that the PGW is not an aggregator. Fixes #87.
- Make clear that the PGW is not an event store. Fixes #89.

@stuartnelson3

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
-# Prometheus Pushgateway [![Build Status](https://travis-ci.org/prometheus/pushgateway.svg)][travis]
+# Prometheus Pushgateway
 
+[![Build Status](https://travis-ci.org/prometheus/pushgateway.svg)][travis]
 [![CircleCI](https://circleci.com/gh/prometheus/pushgateway/tree/master.svg?style=shield)][circleci]
 [![Docker Repository on Quay](https://quay.io/repository/prometheus/pushgateway/status)][quay]
 [![Docker Pulls](https://img.shields.io/docker/pulls/prom/pushgateway.svg?maxAge=604800)][hub]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,19 +11,32 @@ exist long enough to be scraped, they can instead push their metrics
...
+rather a metrics cache. It does not have a statsd-like semantics. The metrics
+pushed are exactly the same as you would present for scraping in a permanently
+running program.
 
 For machine-level metrics, the
 [textfile](https://github.com/prometheus/node_exporter/blob/master/README.md#textfile-collector)
-collector of the Node exporter is usually more appropriate. The Pushgateway is best
-used for service-level metrics.
+collector of the Node exporter is usually more appropriate. The Pushgateway is
+best used for service-level metrics.
```
stuartnelson3: /s/best used/intended for/ ? Purely an opinion, nothing wrong with what is currently written.

README.md
```diff
@@ -10,19 +11,32 @@ exist long enough to be scraped, they can instead push their metrics
...
 
 ## Run it
 
-Compile the binary using the provided Makefile (type `make`).
+Download binary releases for your platform from the
+[release page](https://github.com/prometheus/pushgateway/releases) and unpack
+the tarball.
+
+If you want to compile yourself from the sources, you need a working Go
+setup. Then use the provided Makefile (type `make`).
```
stuartnelson3: Is there a required version?

beorn7: I don't know (and would take time to find out). And it might change any time.

Since this is not the path for regular users, I don't want to invest that effort.

README.md
beorn7: Reads better to myself, too. (The wording here isn't mine.) Will update.

README.md
stuartnelson3: 👍

stuartnelson3:
👍

PR #98, opened 2016-12-06
Support HTTP Basic authentication for Pushgateway access

anarcher:
Sometimes, I need to access the push gateway with HTTP Basic auth. I added HTTP basic auth simply.

Thanks,
MS


grobie:
As there are a huge number of different authentication and authorization means available, the Prometheus project decided to not support any of them and let users fulfill such needs by using reverse proxies in front of the Prometheus server / pushgateway / etc.

beorn7:
Yes, indeed. Sorry @anarcher , we have been planning a non-goal section in the docs, see https://github.com/prometheus/docs/issues/149 , where you can also find server-side HTTP auth in the list.

Sorry again.

anarcher:
I understood it. Thank you for your kind response. :-)

Issue #73, opened 2016-01-26
-web.telemetry-path not honored for pushes


bluecmd:
Hi,

I'm using pushgateway behind a proxy served by /push/*.

For this I run:

```
/srv/pushgateway -web.telemetry-path=/push/metrics
```

This works just fine:

```
$ wget -O - http://localhost:9091/metrics | wc -l
2016-01-26 22:07:15 ERROR 404: Not Found.
0
$ wget -O - http://localhost:9091/push/metrics | wc -l
2016-01-26 22:07:21 (1.20 GB/s) - written to stdout [6777/6777]
118
```

But for pushing this is not the same:

```
$ echo "some_metric 3.14" | curl --data-binary @-  http://localhost:9091/metrics/job/some_job
$
$ echo "some_metric 3.14" | curl --data-binary @-  http://localhost:9091/push/metrics/job/some_job
404 page not found
```

juliusv:
Yeah, `-web.telemetry-path` is only for the normal Prometheus metrics exposition endpoint, not the pushgateway API. However, we could consider supporting `-web.external-url` in the PGW as we do for Prometheus, which would allow you to prefix all path handlers.

csawyerYumaed:
-web.external-url would be nice to have.

PR #112, opened 2017-04-05
adding external-url

joshdvir:
I'm using a reverse proxy using path so I need the external url same as prometheus.

When using the external url it brakes the template.html since it can't reach the static files, I don't have any idea's how to fix it.

#73

### DIFF
--- a/main.go
+++ b/main.go
@@ -69,36 +69,40 @@ func main() {
 	// prometheus.EnableCollectChecks(true)
 
 	r := httprouter.New()
+	if *prefixPath != "" {
+		r = r.WithPrefix(*prefixPath)
+	}
+
 	r.Handler("GET", *metricsPath, prometheus.Handler())
 
 	// Handlers for pushing and deleting metrics.
-	r.PUT(*prefixPath+"/metrics/job/:job/*labels", handler.Push(ms, true))
-	r.POST(*prefixPath+"/metrics/job/:job/*labels", handler.Push(ms, false))
-	r.DELETE(*prefixPath+"/metrics/job/:job/*labels", handler.Delete(ms))
-	r.PUT(*prefixPath+"/metrics/job/:job", handler.Push(ms, true))
-	r.POST(*prefixPath+"/metrics/job/:job", handler.Push(ms, false))
-	r.DELETE(*prefixPath+"/metrics/job/:job", handler.Delete(ms))
+	r.PUT("/metrics/job/:job/*labels", handler.Push(ms, true))
+	r.POST("/metrics/job/:job/*labels", handler.Push(ms, false))
+	r.DELETE("/metrics/job/:job/*labels", handler.Delete(ms))
+	r.PUT("/metrics/job/:job", handler.Push(ms, true))
+	r.POST("/metrics/job/:job", handler.Push(ms, false))
+	r.DELETE("/metrics/job/:job", handler.Delete(ms))
 
 	// Handlers for the deprecated API.
-	r.PUT(*prefixPath+"/metrics/jobs/:job/instances/:instance", handler.LegacyPush(ms, true))
-	r.POST(*prefixPath+"/metrics/jobs/:job/instances/:instance", handler.LegacyPush(ms, false))
-	r.DELETE(*prefixPath+"/metrics/jobs/:job/instances/:instance", handler.LegacyDelete(ms))
-	r.PUT(*prefixPath+"/metrics/jobs/:job", handler.LegacyPush(ms, true))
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -38,6 +38,7 @@ var (
 	showVersion         = flag.Bool("version", false, "Print version information.")
 	listenAddress       = flag.String("web.listen-address", ":9091", "Address to listen on for the web interface, API, and telemetry.")
 	metricsPath         = flag.String("web.telemetry-path", "/metrics", "Path under which to expose metrics.")
+	prefixPath          = flag.String("web.external-url", "", "The URL under which PushGateway is externally reachable.")
```
beorn7: Further down, we need to sanitize the flag value in the way it is done [here](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/config.go#L300).

Nit: Please don't CamelCase Pushgateway. I suggest "...under which the Pushgateway is externally reachable." This is consistent with the wording and capitalization used in the README.md.

main.go
```diff
@@ -38,6 +38,7 @@ var (
 	showVersion         = flag.Bool("version", false, "Print version information.")
 	listenAddress       = flag.String("web.listen-address", ":9091", "Address to listen on for the web interface, API, and telemetry.")
 	metricsPath         = flag.String("web.telemetry-path", "/metrics", "Path under which to expose metrics.")
+	prefixPath          = flag.String("web.external-url", "", "The URL under which Pushgateway is externally reachable.")
```
beorn7: Nit nit: "... which the Pushgateway..."

main.go
```diff
@@ -71,33 +72,33 @@ func main() {
 	r.Handler("GET", *metricsPath, prometheus.Handler())
 
 	// Handlers for pushing and deleting metrics.
-	r.PUT("/metrics/job/:job/*labels", handler.Push(ms, true))
-	r.POST("/metrics/job/:job/*labels", handler.Push(ms, false))
-	r.DELETE("/metrics/job/:job/*labels", handler.Delete(ms))
-	r.PUT("/metrics/job/:job", handler.Push(ms, true))
-	r.POST("/metrics/job/:job", handler.Push(ms, false))
-	r.DELETE("/metrics/job/:job", handler.Delete(ms))
+	r.PUT(*prefixPath+"/metrics/job/:job/*labels", handler.Push(ms, true))
```
beorn7: We have a special wrapped router to do this in the Prometheus server, see https://github.com/prometheus/prometheus/blob/master/web/web.go#L167 . I'd suggest to do it in the same way here to avoid the many `*prefixPath` replications.

beorn7:
Thanks for doing this. See some comments above.

main.go
joshdvir: Fixed the name.

The link you sent wasn't sanitation it was utilization of a different variable matching the value.

main.go
joshdvir: I'm missing something here, can you please write the whole sentence

main.go
beorn7: Sorry, that's two lines further down: https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/config.go#L303

joshdvir:
@beorn7 I think it looks good now, what do you think ?

main.go
beorn7: "The URL under which the Pushgateway is externally reachable."

(Article "the" is missing. In README.md, we usually call it "the Pushgateway".)

beorn7:
Looks good (modulo article nit). Let's wait for the CI to go green.

Thanks again.

joshdvir:
There is a problem there with the WithPrefix function since in the Pushgateway you use directly with the router and the Prometheus wrapper, I'm trying to use the Prometheus wrapper now.

beorn7:
@joshdvir Are you still working on this?

joshdvir:
Hi @beorn7,

Sadly I didn't find time to finish this, not sure when I will make it.

Josh.

beorn7:
No worries. I'll take inspiration from your PR then and finish it up myself soon.

hairyhenderson:
@beorn7 any progress on this? 🙂

beorn7:
This got preempted by higher priority things. If you want to take a stab, be my guest. Otherwise, I'll keep it on my list.

hairyhenderson:
@beorn7 thanks for the quick response! I'll give it a shot, but not sure I have enough spare time to complete it...

hairyhenderson:
@beorn7 FYI - I've submitted #146 for this

beorn7:
Superseded by #146

PR #116, opened 2017-04-20
Point at users list rather than dev

brian-brazil:


resources/template.html
```diff
@@ -219,7 +219,7 @@ <h4 class="panel-title">
     <footer>
       <p>
         Code: <a href="https://github.com/prometheus/pushgateway">prometheus/pushgateway</a>
-        Contact: <a href="mailto:prometheus-developers@googlegroups.com">prometheus-developers</a>
```
grobie: What about just linking to our community page, so that we only have to care about one canonical page?

brian-brazil:
Reworked it to make that work.

grobie:
:+1:

brian-brazil:
@beorn7 Do you have any comments on this?

IIRC this is here only so it doesn't look like the homepage an empty pushgateway isn't stuck loading, so the content doesn't matter much.

PR #119, opened 2017-05-24
Reject pushes with timestamps.

### LINKED ISSUE #109: Don't accept timestamps from pushes
My ruminations on https://docs.google.com/document/d/1ordTPfUSaGvaBolUGeLChATwgc1xUhUxfUhudYoOeMw/edit indicate that my original plan of having the pushgateway ultimately expose data with the timestamp it was pushed at is unworkable.

Given this, it appears there are no valid use cases for timestamps with the pushgateway. Accordingly I propose we either silently drop or hard reject pushes containing timestamps, as everyone we've seen doing this so far was using the pgw in an improper fashion.

Once this is done, I think it'd be safe to client libraries to support timestamps for the handful of advanced users who need them. Up to now, (almost) everyone asking for timestamps in client libraries was trying to incorrectly use them with the pushgateway - so if we make that impossible then the abuse risk is as mitigated as it's going to get.

brian-brazil:
Fixes #109

I once thought that some day we'd solve staleness, and there
would be valid use cases for the pushgateway to expose timestamps.
With the recent work on staleness this has turned out not to be the case.
As every single person we have seen pushing timestamps to the pgw
has been doing so incorrectly (usually misguidedly trying to make
Prometheus do push), let's send a stronger signal that this is not the
right way to do things.

With this in place I think we'll have all the mitigation we're going to
get in terms of discourage users from misusing timestamps, and it'll be
tolerable to add timestamp support to client libraries for the very very
few use cases that need this advanced feature (at current count:
collectd, cloudwatch, graphite and inflxudb exporters).

This should in principle have no impact as a) there is no documentation
saying that this is a good idea or even an idea and b) there is no code
support in our libraries for pushing timestamps.

@beorn7 @fabxc @juliusv 

https://github.com/prometheus/prometheus/issues/398 is going to be resolved very soon. I've always said that staleness blocked adding timestamps to clients due to repeated abuse by users via the pgw (despite us not offering any library support for such abuse), however it turns out staleness handling doesn't help. So let's do this instead to discourage the primary method of abuse, and once this is released we can unblock client library timestamps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,6 @@ Examples:
         cat <<EOF | curl --data-binary @- http://pushgateway.example.org:9091/metrics/job/some_job/instance/some_instance
         # TYPE some_metric counter
         some_metric{label="val1"} 42
-        # This one even has a timestamp (but beware, see below).
-        some_metric{label="val2"} 34 1398355504000
         # TYPE another_metric gauge
         # HELP another_metric Just an example.
         another_metric 2398.283
@@ -157,11 +155,12 @@ be scraped at all anymore. (Prometheus knows only one timestamp per
 sample, there is no way to distinguish a 'time of pushing' and a 'time
 of scraping'.)
 
-You can still force Prometheus to attach a different timestamp by
-using the optional timestamp field in the exchange format. However,
-there are very few use cases where that would make
-sense. (Essentially, if you push more often than every 5min, you
-could attach the time of pushing as a timestamp.)
+As there are essentially no use cases where it would make sense to to attach a
+different timestamp, and many users attempting to incorrectly do so (despite no
+client library supporting this) any pushes with timestamps will be rejected.
+
+If you think you need to push a timestamp, please see [When To Use The
+Pushgateway](https://prometheus.io/docs/practices/pushing/).
 
 ## API
 
--- a/handler/handler_test.go
+++ b/handler/handler_test.go
@@ -195,6 +195,30 @@ func TestPush(t *testing.T) {
 		t.Errorf("Wanted metric family %v, got %v.", expected, got)
 	}
 
+	// With job name and instance name and timestamp specified.
+	mms.lastWriteRequest = storage.WriteRequest{}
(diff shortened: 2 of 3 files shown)


beorn7:
I know that a number of people push with timestamp, even if it is for either plain abuse or at least something very hacky (emulating bulk import, which we still don't offer natively). "No impact" is thus not really the right wording.

Personally, I agree with the change. People who depend on their hacky solution could just continue to use an older version of the PGW for it. Code looks good. I would still prefer to get other opinions than just my own.

beorn7:
And I'll take silence as consent. :)

If nobody comments on this in the next couple of days, let's just merge.

juliusv:
:+1: Sounds good to me too.

brian-brazil:
> "No impact" is thus not really the right wording.

I said "in principle" for a reason :)

beorn7:
Merging in 3… 2… 1…

jomach:
I think this was a wrong move. This was the only way to insert Historical data... My use case: 
We have metrics on opentsdb and we want to start using prometheus api to query the data, so I need to migrate the data from opentsdb to prometheus. Guess what... it is not possible which kind of  s****

juliusv:
@jomach FWIW that use case (basically turning Prometheus into a general push-based system) is explicitly not what the Pushgateway was made for, see https://prometheus.io/docs/practices/pushing/.

However, you might be interested in https://github.com/prometheus/prometheus/issues/535 about bulk imports, an issue that has been picked up again recently and will likely happen soon.

beorn7:
@jomach Sorry for the disappointment. But as discussed above, it's indeed deliberate that we don't want to cater for the use case of turning the PGW into a (very very brittle) backfill engine. As also mentioned [above](https://github.com/prometheus/pushgateway/pull/119#issuecomment-304611655), the old versions of the PGW are not out of the world. You can still download them if you insist to keep running it as you did.

Also, as you have seen now, there is some progress to be expected on native bulk import in the not too far future.

PR #120, opened 2017-05-31
Add push_time_seconds for when the last push was.

### LINKED ISSUE #108: Add a synthetic metric indicating when the last push happens
I propose that all groups should automatically expose a metric called "push_time_seconds" or similar indicating when the last push/pushadd happened. This would have the labels of its respective group.

As making the timestamp of pushed data be the timestamp of the push looks like it's not going to work out, this offers another way to tell if a push hasn't happened recently enough.

brian-brazil:
Fixes #108 

@beorn7 

I'm not fully happy with the metric name, but it is consistent with what's already in the UI - we don't currently distinguish between push and pushadd there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ time. A metric that cannot be scraped has basically ceased to
 exist. Prometheus is somewhat tolerant, but if it cannot get any
 samples for a metric in 5min, it will behave as if that metric does
 not exist anymore. Preventing that is actually one of the reasons to
-use a push gateway. The push gateway will make the metrics of your
+use a Pushgateway. The Pushgateway will make the metrics of your
 ephemeral job scrapable at any time. Attaching the time of pushing as
 a timestamp would defeat that purpose because 5min after the last
 push, your metric will look as stale to Prometheus as if it could not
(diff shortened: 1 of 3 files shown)


handler/push.go
```diff
@@ -295,3 +304,19 @@ func timestampsPresent(metricFamilies map[string]*dto.MetricFamily) bool {
 	}
 	return false
 }
+
+// Checks if any timestamps have been specified.
```
beorn7: Doc comment here is a left over.

brian-brazil:
Updated, PTAL.

README.md
```diff
@@ -162,6 +162,11 @@ client library supporting this) any pushes with timestamps will be rejected.
 If you think you need to push a timestamp, please see [When To Use The
 Pushgateway](https://prometheus.io/docs/practices/pushing/).
 
+In order to make it easier to alert on pushers that have not run recently, the
+push gateway will add in a metric `push_time_seconds` with the unix timestamp
```
beorn7: We use the spelling "Pushgateway" in this file.

beorn7: And now that we are concerned with spelling nits: I'd capitalize Unix. Cf. https://en.wikipedia.org/wiki/Unix_time .

handler/push.go
```diff
@@ -36,6 +36,11 @@ import (
 	"github.com/prometheus/pushgateway/storage"
 )
 
+const (
+	pushMetricName = "push_time_seconds"
+	pushMetricHelp = "Last unixtime when this group was changed in the pushgateway."
```
beorn7: In the spirit of the above, I suggest "Last Unix time when this group was changed in the Pushgateway."

README.md
brian-brazil: Done, and the others in the file.

handler/push.go
brian-brazil: Done

beorn7:
Thanks for consistency.

PR #123, opened 2017-06-09
Debug level logging now shows client-induced errors

### LINKED ISSUE #55: Improve debug logging
Currently, we don't use leveled logging at all in pushgateway.

The original idea was that the pushgateway is so simplistic, all the relevant information would be in the HTTP response.

As it turns out, clients tend to never display the body of an HTTP response with an error code.

For convenience, consider introducing leveled logging so that client-induced errors (invalid payload) are logged at a low level like INFO or DEBUG.

Conorbro:
Implements #55

### DIFF
--- a/handler/delete.go
+++ b/handler/delete.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
 
 	"github.com/prometheus/pushgateway/storage"
 )
@@ -41,10 +42,12 @@ func Delete(ms storage.MetricStore) func(http.ResponseWriter, *http.Request, htt
 			labels, err := splitLabels(labelsString)
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusBadRequest)
+				log.Debugf("Failed to parse URL: %v, %v", labelsString, err.Error())
 				return
 			}
 			if job == "" {
 				http.Error(w, "job name is required", http.StatusBadRequest)
+				log.Debug("job name is required")
 				return
 			}
 			labels["job"] = job
@@ -79,6 +82,7 @@ func LegacyDelete(ms storage.MetricStore) func(http.ResponseWriter, *http.Reques
 
 			if job == "" {
 				http.Error(w, "job name is required", http.StatusBadRequest)
+				log.Debug("job name is required")
 				return
 			}
 			labels := map[string]string{"job": job}
--- a/handler/push.go
+++ b/handler/push.go
@@ -29,6 +29,7 @@ import (
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
 
 	dto "github.com/prometheus/client_model/go"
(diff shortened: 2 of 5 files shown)


handler/delete.go
```diff
@@ -41,10 +42,12 @@ func Delete(ms storage.MetricStore) func(http.ResponseWriter, *http.Request, htt
 			labels, err := splitLabels(labelsString)
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusBadRequest)
+				log.Debug(err.Error(), http.StatusBadRequest)
```
brian-brazil: This could be a bit hard to understand among other logs, I'd suggest formatting these with "Failed to parse URL %v: %v" etc.

handler/delete.go
```diff
@@ -41,10 +42,12 @@ func Delete(ms storage.MetricStore) func(http.ResponseWriter, *http.Request, htt
 			labels, err := splitLabels(labelsString)
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusBadRequest)
+				log.Debugf("Failed to parse URL, %v: %v", err.Error(), http.StatusBadRequest)
```
brian-brazil: I'd include the URL, no need for the return code.

And below.

storage/diskmetricstore.go
```diff
@@ -149,7 +149,7 @@ func (dms *DiskMetricStore) loop(persistenceInterval time.Duration) {
 					persistStarted := time.Now()
 					if err := dms.persist(); err != nil {
 						log.Errorln("Error persisting metrics:", err)
-					} else {
+					} else if len(dms.persistenceFile) > 0 {
```
brian-brazil: We still persisted if there was nothing.

Conorbro:
@brian-brazil

Conorbro:
@brian-brazil

handler/push.go
```diff
@@ -102,10 +105,12 @@ func Push(
 			}
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusBadRequest)
+				log.Debug("Failed to parse text, %v: %v", err.Error())
```
brian-brazil: You only have one argument here

handler/push.go
```diff
@@ -62,10 +63,12 @@ func Push(
 			labels, err := splitLabels(labelsString)
 			if err != nil {
 				http.Error(w, err.Error(), http.StatusBadRequest)
+				log.Debug("Failed to parse URL, %v: %v", err.Error())
```
brian-brazil: There's only one argument here, and using % requires Debugf

Conorbro:
@brian-brazil

handler/status.go
```diff
@@ -59,11 +60,13 @@ func Status(
 		tpl, err := assetFunc("template.html")
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
+			log.Warnf("Error loading template.html, %v", err.Error())
```
brian-brazil: This should be an error, this should never happen.

handler/status.go
```diff
@@ -59,11 +60,13 @@ func Status(
...
 			return
 		}
 		_, err = t.Parse(string(tpl))
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
+			log.Debugf("Error parsing template, %v", err.Error())
```
brian-brazil: Ditto

beorn7:
Will have a look ASAP. Hopefully tomorrow.

beorn7:
:+1: Looks good. Thanks.

PR #125, opened 2017-07-30
To be able to compile it on arm architecture too

sdorin:
Otherwise it fails with messages like:
`compile: unknown architecture "armv7l"`


beorn7:
Hi @sdorin , my apologies for the delay. I went on vacations and missed to delegate my maintainer duties to somebody else.

This looks fine to me. I guess we need to propagate this to other repos, too. @sdurrheimer is that true or is there a better way?

Issue #124, opened 2017-07-28
Consider responding with status code 400 to PUT and POST with empty content.


nordbranch:
I have one group who's pushing metrics to a pushgateway, but their Group was mysteriously disappearing.  I tcpdumped the traffic to catch if someone had setup a DELETE call somewhere, but didn't find any such traffic.  I then noticed this (using "foobar" for redaction):

PUT /metrics/job/foobar HTTP/1.1
Content-Type: text/plain; version=0.0.4; charset=utf-8
User-Agent: Java/1.8.0_74
Host: foobar:9091
Accept: text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2
Connection: keep-alive
Content-Length: 0

Note the "Content-Length: 0".  Turns out this will cause the Group to be deleted.  I'm not sure if this should be allowed, or if this should return an error (in favor of forcing use of DELETE), but thought I'd report it in for discussion/consideration.

beorn7:
I would say that's "working as expected".

It kind of makes me think we never really needed an explicit `DELETE` verb.

Happy to add a guard returning 400 for empty `PUT` or `POST`, but as you say, it would be good to see more opinions.

brian-brazil:
I think the current behaviour is correct, as it's consistent.

beorn7:
Without doubt, the current behavior is consistent, and also technically correct (and as we all know, technically correct is the best kind of correct).

The argument was not so much about consistency but about avoiding confusion if an empty body is sent accidentally. I'd assume that nobody has ever sent an empty body on purpose. Thus, while there would fundamentally be a loss of consistency if we rejected empty bodies, it would be mostly irrelevant.

Personally, I see both points and would lean slightly towards not changing the status quo. I take @brian-brazil 's statement as preferring the current consistent behavior. If nobody chimes in with a different opinion, I'll close this as "works as intended" in a couple of days.

beorn7:
OK, we'll leave it as it is. #126 to clarify the docs.

PR #126, opened 2017-08-10
Improve README.md

beorn7:
This addresses the confusion in #124.

@brian-brazil @nordbranch

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,10 @@ Prometheus.
 ## Non-goals
 
 The Pushgateway is explicitly not an _aggregator or distributed counter_ but
-rather a metrics cache. It does not have a statsd-like semantics. The metrics
-pushed are exactly the same as you would present for scraping in a permanently
-running program.
+rather a metrics cache. It does not have
+[statsd](https://github.com/etsy/statsd)-like semantics. The metrics pushed are
+exactly the same as you would present for scraping in a permanently running
+program.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,9 +14,10 @@ Prometheus.
 ## Non-goals
 
 The Pushgateway is explicitly not an _aggregator or distributed counter_ but
-rather a metrics cache. It does not have a statsd-like semantics. The metrics
-pushed are exactly the same as you would present for scraping in a permanently
-running program.
+rather a metrics cache. It does not have a
```
brian-brazil: "It does not have statsd-like..." would read better

README.md
```diff
@@ -250,12 +251,19 @@ processed. Neither is there a guarantee that the pushed metrics are
 persisted to disk. (A server crash may cause data loss. Or the push
 gateway is configured to not persist to disk at all.)
 
+A `PUT` request with an empty body effectively deletes all metrics with the
```
brian-brazil: Hmm, shouldn't that set push_time_seconds still?

README.md
beorn7: FIxed.

README.md
beorn7: Indeed. So it's not strictly equivalent... Joy. I updated the wording here.

beorn7:
Thanks for reviewing.

PR #127, opened 2017-08-14
Move docker instructions to the Run-it section.

metalmatze:
I was looking for instructions on how to run with docker. It was interesting to find this at the very bottom of the README. I think the Run-it section is better suited.


beorn7:
Sure, thanks.

Issue #105, opened 2017-02-22
Provide a HTTP Health Check


iksaif:
Kubernetes, Marathon, consul and others require health checks, and often an HTTP based one. While Prometheus and Alertmanager can somewhat use their status page this is quite painful with pushgateway because it contains all the metrics that have been pushed to it.

Could we consider adding either a separate light status page or a dedicated /health endpoint ?

beorn7:
That's a good point. Personally, I don't have much experience with running those stateful Prometheus components on Kubernetes, so I didn't really run into it myself.

@fabxc what's your take on this topic? You and your CoreOS colleagues should be much more directly affected. Would it be a good idea to provide a dedicated health endpoint on all our components? This could involve some real health checking where appropriate (which is something the status page doesn't provide).

jsravn:
This would be really useful. Our status page is 6M (don't ask why! :)) and the metrics endpoint is 2M. This means we need quite long probe timeouts.

beorn7:
Pinging @fabxc again (as I want to implement this in a way that will be useful across components of the project, not just for this special case).
@fabxc : How do you do health checking for Prometheus components in K8s so far? How do you do liveness vs. readiness probes? Should we perhaps support both with an endpoint?

fabxc:
Currently we also hit `/status` on Prometheus and check for 200.
A dedicated health endpoint would probably be good. Can just return 200 initially but it will enable use to add some internal checks under the hood later.

That'd be liveness probe.

Readiness could be added, e.g. waiting for Prometheus to load the initial checkpoint. But I'd only add endpoints once we have something to actually check. Other components probably don't have such a condition.

beorn7:
Currently, Prometheus doesn't serve any HTTP while loading the checkpoint (or worse: running crash recovery). Definitely something to think about.

Are there established conventions for the path of those liveness/readiness endpoints?

fabxc:
Kubernetes components use /healthz for reasons you can imagine. In general
there are no defaults or conventions beyond that.

beorn7:
I'm _soooo_ tempted to use `/livenezz` and `/readinezz` now… :)

If there are no conventions yet, I'd suggest `/health` and `/readiness` (nouns) or `/healthy` and `/ready` (adjectives).

[...]

PR #135, opened 2017-10-09
Add /-/ready and /-/healty

iksaif:
Also add the associated functions to one of the
most important sub-system of the pushgateway: metric store.

The http handler will currently mostly reflect the status of
the metric store, but there isn't much else that can break.

Since the current implementation of the disk metric store does
most of the init job in its constructor, /-/ready isn't super
useful yet.

See https://github.com/prometheus/pushgateway/issues/105

### DIFF
--- a/handler/misc.go
+++ b/handler/misc.go
@@ -1,0 +1,47 @@
+// Copyright 2017 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package handler
+
+import (
+	"io"
+	"net/http"
+
+	"github.com/prometheus/pushgateway/storage"
+)
+
+func Healthy(
+	ms storage.MetricStore,
+) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		err := ms.Healthy()
+		if err == nil {
+			io.WriteString(w, "OK")
+		} else {
+			http.Error(w, err.Error(), 500)
+		}
+	}
+}
+
+func Ready(
+	ms storage.MetricStore,
+) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		err := ms.Ready()
+		if err == nil {
+			io.WriteString(w, "OK")
+		} else {
+			http.Error(w, err.Error(), 500)
+		}
+	}
+}
--- a/storage/diskmetricstore.go
+++ b/storage/diskmetricstore.go
@@ -15,6 +15,7 @@ package storage
 
 import (
 	"encoding/gob"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -61,6 +62,8 @@ func NewDiskMetricStore(
 	persistenceFile string,
 	persistenceInterval time.Duration,
 ) *DiskMetricStore {
+	// TODO: Do that outside of the constructor to allow the HTTP server to
+	//  serve /-/healthy and /-/ready earlier.
 	dms := &DiskMetricStore{
 		writeQueue:      make(chan WriteRequest, writeQueueCapacity),
 		drain:           make(chan struct{}),
@@ -132,6 +135,26 @@ func (dms *DiskMetricStore) GetMetricFamilies() []*dto.MetricFamily {
 func (dms *DiskMetricStore) Shutdown() error {
 	close(dms.drain)
 	return <-dms.done
+}
+
+// Healthy implements the MetricStore interface.
+func (dms *DiskMetricStore) Healthy() error {
+	// By taking the lock we check that there is no deadlock.
+	dms.lock.Lock()
+	defer dms.lock.Unlock()
+
+	// A pushgateway that cannot be written to should not be
+	// considered as healthy.
+	if len(dms.writeQueue) == cap(dms.writeQueue) {
+		return fmt.Errorf("write queue is full")
+	}
+
+	return nil
+}
+
+// Ready implements the MetricStore interface.
+func (dms *DiskMetricStore) Ready() error {
+	return dms.Healthy()
 }
 
 func (dms *DiskMetricStore) loop(persistenceInterval time.Duration) {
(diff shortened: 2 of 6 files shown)


handler/misc.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2014 The Prometheus Authors
```
beorn7: It's legally (almost) irrelevant, but I'd put the current year into newly created files.

iksaif: done

iksaif:
Right, better now

beorn7:
Looking…

storage/interface.go
```diff
@@ -56,6 +56,13 @@ type MetricStore interface {
 	// undefinded state). If nil is returned, the MetricStore cannot be
 	// "restarted" again, but it can still be used for read operations.
 	Shutdown() error
+	// Returns true, nil if the MetricStore is currently working as expected
+	// or false, Error if it is not.
+	Healthy() (bool, error)
```
beorn7: Couldn't we just return `error` with the usual sentinel semantics that `nil` is OK and everything else is an error?

storage/interface.go
```diff
@@ -56,6 +56,13 @@ type MetricStore interface {
 	// undefinded state). If nil is returned, the MetricStore cannot be
 	// "restarted" again, but it can still be used for read operations.
 	Shutdown() error
+	// Returns true, nil if the MetricStore is currently working as expected
```
beorn7: Please start the doc comment with `Healthy`, as above, i.e. `// Healthy returns true...`.

storage/interface.go
```diff
@@ -56,6 +56,13 @@ type MetricStore interface {
...
+	// or false, Error if it is not.
+	Healthy() (bool, error)
+	// Returns true, nil if the MetricStore is ready to be used (all files
+	// are opened and checkpoints have been restored) or false, Error if it
+	// is not.
+	Ready() (bool, error)
```
beorn7: Same comments apply here.

storage/diskmetricstore.go
```diff
@@ -134,6 +137,24 @@ func (dms *DiskMetricStore) Shutdown() error {
 	return <-dms.done
 }
 
+func (dms *DiskMetricStore) Healthy() (bool, error) {
```
beorn7: Need a doc comment, something like `// Healthy implements the MetricStore interface.` will suffice.

storage/diskmetricstore.go
```diff
@@ -134,6 +137,24 @@ func (dms *DiskMetricStore) Shutdown() error {
...
+	// A pushgateway that cannot be written to should not be
+	// considered as healthy.
+	if len(dms.writeQueue) == cap(dms.writeQueue) {
+		return false, fmt.Errorf("Write queue is full")
+	}
+
+	return true, nil
+}
+
+func (dms *DiskMetricStore) Ready() (bool, error) {
```
beorn7: Doc comment missing, see above.

handler/misc.go
```diff
@@ -0,0 +1,63 @@
...
+		func(w http.ResponseWriter, _ *http.Request) {
+			healthy, err := ms.Healthy()
+			if healthy {
+				io.WriteString(w, "OK")
+			} else {
+				fmt.Fprintf(w, "FAIL: %v", err)
+			}
+		},
+	)
+	return func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
```
beorn7: Since you are not using the httprouter.Params, you can do this in an easier way, see the `Status` handler.

handler/misc.go
```diff
@@ -0,0 +1,63 @@
...
+		func(w http.ResponseWriter, _ *http.Request) {
+			ready, err := ms.Ready()
+			if ready {
+				io.WriteString(w, "OK")
+			} else {
+				fmt.Fprintf(w, "FAIL: %v", err)
+			}
+		},
+	)
+	return func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
```
beorn7: See above.

handler/misc.go
```diff
@@ -0,0 +1,63 @@
...
+func HealthyHandler(
+	ms storage.MetricStore,
+) func(http.ResponseWriter, *http.Request, httprouter.Params) {
+	handlerFunc := prometheus.InstrumentHandlerFunc("healthy",
+		func(w http.ResponseWriter, _ *http.Request) {
+			healthy, err := ms.Healthy()
+			if healthy {
+				io.WriteString(w, "OK")
+			} else {
+				fmt.Fprintf(w, "FAIL: %v", err)
```
beorn7: In this case, you also want to set the HTTP status code to 500. Cf. https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-http-request

storage/interface.go
iksaif: done

storage/interface.go
iksaif: good idea

storage/interface.go
iksaif: done

storage/diskmetricstore.go
iksaif: done

storage/diskmetricstore.go
iksaif: done

iksaif:
Fixed part of the issues locally. I'll finish the changes and push the new code during the day.

handler/misc.go
iksaif: done

iksaif:
> In general, it would be great if you could run go vet and golint on your code to see any problems.

Sure, will do, I half-expected that to be done in the presbumit, but I'll add myself a few aliases

iksaif:
Updated !

beorn7:
> Sure, will do, I half-expected that to be done in the presbumit, but I'll add myself a few aliases

It is in the better maintained repositories. Pushgateway is a smaller one with not so much activity (which is of course no excuse to not do it here, too).

Thanks again, will have a look ASAP.

storage/diskmetricstore.go
```diff
@@ -134,6 +137,26 @@ func (dms *DiskMetricStore) Shutdown() error {
...
+// Healthy implements the MetricStore interface.
+func (dms *DiskMetricStore) Healthy() error {
+	// By taking the lock we check that there is no deadlock.
+	dms.lock.Lock()
+	defer dms.lock.Unlock()
+
+	// A pushgateway that cannot be written to should not be
+	// considered as healthy.
+	if len(dms.writeQueue) == cap(dms.writeQueue) {
+		return fmt.Errorf("Write queue is full")
```
beorn7: Final nit: error strings should not be capitalized.

beorn7:
Looks great. Once the error nit is done, I'll merge.

iksaif:
done

beorn7:
Thanks once more. Merging…

tinproject:
There is a typo on the name of this PR, it saids __healty__ instead of __healthy__. 

It's also on the releases changelog for the 0.5.0 release. I do not know if it worths changing it to avoid mistakes or not.

beorn7:
Changed here, and will fix the CHANGELOG.md, too. thanks.

tinproject:
Still on the Github releases page: https://github.com/prometheus/pushgateway/releases/tag/v0.5.0

beorn7:
Fixed now, too.

Issue #143, opened 2018-02-07
Cut 0.5.0


beorn7:
Requested by some...
#90 would be really good to get fixed before.

beorn7:
And #101, as this is another breaking change to the persistence format, and should be done in one go.

Issue #90, opened 2016-08-29
Gob'ing protobufs goes wrong in subtle ways


beorn7:
In contrast to what might or might not have been my assumptions back then, if you use `gob` to encode protobuf messages, the normal marshaling and unmarshaling of protobuf is sidestepped (unsurprisingly, in hindsight…).

Here are the issues:

`gob` encounters a proto message and sees it as the struct it is in Go. That struct has pointers as fields, and if a protobuf field is unset, the corresponding pointer is `nil`. Naively read, the `gob` documentation tells you that a `nil` pointer cannot be encoded and will lead to a panic or error. For some reason, however, the implementation tolerates `nil` pointers in structs and simply ignores them during encoding. This combines in an interesting fashion with another `gob` feature: If a field in a struct has its zero value (including the case if a pointer points to a zero value), `gob` will not encode it (as an optimization, I assume). During decoding of pointer fields in a struct, however, both cases (`nil` pointer as well as pointer to a zero value) will result in a `nil` pointer. (Both have been ignored during encoding.) That has the interesting side effect that any protobuf fields that happen to correspond to a Go zero value of their type, will come out as "unset" (in protobuf lingo) from the `gob` decoding.

There might be more issues with gob`ing protobuf messages. The original sin here is to gob protobuf messages in the first place. This has to be fixed (by introducing an intermediate interface that will use the protobuf Marshal and Unmarshal methods), which will unfortunately change the storage format once more, so some kind of transparent upgrade needs to be provided.

PR #146, opened 2018-02-23
Add web.route-prefix flag

### LINKED ISSUE #73: -web.telemetry-path not honored for pushes

hairyhenderson:
Fixes #73 and supersedes #112 

I took a different approach - somewhat less elegant than originally proposed in #112. I'm open to suggestions 😉

I named the flag `-web.route-prefix` instead of `-web.external-url`, since it seemed the former is closer to what I think we're looking for in terms of how they're defined already in `prometheus`.

An aside - I had to dig into the git history to figure out how to do the `bindata.go` generation - seems the `bindata.go` Makefile target got removed in a previous commit, not sure if that was intentional. The command I used was:

```console
$ go-bindata -prefix=resources resources/...
```

Let me know how this looks - I'm happy to make any tweaks required!


beorn7:
On my review queue.

beorn7:
Apologies for the delay. I had a busy week. I'm on it now.

beorn7:
BTW: The bindata Make target is now called `assets`. But I think you did the right thing. Essentially what `make assets` would have done anyway. (If you run it, you always get some changes because the file timestamps are baked into the bindata.)

main.go
```diff
@@ -38,6 +38,7 @@ var (
 	showVersion         = flag.Bool("version", false, "Print version information.")
 	listenAddress       = flag.String("web.listen-address", ":9091", "Address to listen on for the web interface, API, and telemetry.")
 	metricsPath         = flag.String("web.telemetry-path", "/metrics", "Path under which to expose metrics.")
+	routePrefix         = flag.String("web.route-prefix", "", "Prefix for the internal routes of web endpoints.")
```
beorn7: Yes, definitely a good choice as it is in line with the current flag of the Prometheus server.

main.go
```diff
@@ -49,13 +50,21 @@ func init() {
 func main() {
 	flag.Parse()
 
+	if *routePrefix == "/" {
+		*routePrefix = ""
+	}
+	if *routePrefix != "" && (*routePrefix)[0:1] != "/" {
+		log.Fatalf("-web.route-prefix value must begin with `/` (was `%s`)", *routePrefix)
+	}
```
beorn7: This requirement is inconsistent with the Prometheus server, which silently prepends the `/`. I think we can do the same here. Suggestion:

```Go
if *routePrefix != "" {
	*routePrefix = "/" + strings.Trim(*routePrefix, "/")
}
```

main.go
```diff
@@ -73,30 +82,38 @@ func main() {
 	r.Handler("GET", *metricsPath, prometheus.Handler())
 
 	// Handlers for pushing and deleting metrics.
-	r.PUT("/metrics/job/:job/*labels", handler.Push(ms, true))
-	r.POST("/metrics/job/:job/*labels", handler.Push(ms, false))
-	r.DELETE("/metrics/job/:job/*labels", handler.Delete(ms))
-	r.PUT("/metrics/job/:job", handler.Push(ms, true))
-	r.POST("/metrics/job/:job", handler.Push(ms, false))
-	r.DELETE("/metrics/job/:job", handler.Delete(ms))
+	pushAPIPath := *routePrefix + "/metrics"
```
beorn7: At this point, we have already registered "/-/healthy", "/-/ready", and "/metrics". I think we want them prefixed as well. (At least that's what the Prometheus server is doing.) WDYT?

main.go
```diff
@@ -73,30 +82,38 @@ func main() {
...
+	r.DELETE(pushAPIPath+"/jobs/:job", handler.LegacyDelete(ms))
+
+	r.Handler("GET", *routePrefix+"/static/*filepath", prometheus.InstrumentHandler(
 		"static",
 		http.FileServer(
-			&assetfs.AssetFS{Asset: Asset, AssetDir: AssetDir, AssetInfo: AssetInfo},
+			&assetfs.AssetFS{
+				Asset: func(name string) ([]byte, error) {
+					// Note: this will do bad things if routePrefix is multi-byte,
+					// but I don't think that's valid in a URL anyway.
```
beorn7: That's actually fine. `len` on a string acts on bytes, not runes, and the slice operations also act on the byte level.

main.go
```diff
@@ -73,30 +82,38 @@ func main() {
...
 		),
 	))
 	statusHandler := prometheus.InstrumentHandlerFunc("status", handler.Status(ms, Asset, flags))
-	r.Handler("GET", "/status", statusHandler)
-	r.Handler("GET", "/", statusHandler)
+	r.Handler("GET", *routePrefix+"/status", statusHandler)
+	r.Handler("GET", *routePrefix+"/", statusHandler)
 
 	// Re-enable pprof.
 	r.GET("/debug/pprof/*pprof", handlePprof)
```
beorn7: We probably want the routing prefix here as well?

beorn7:
About the `-web.external-url`: I don't think we need that in Pushgateway because we do not create any absolute URLs containing the host etc.

hairyhenderson:
> However, the "github.com/prometheus/common/route" seems to have solved the problem of a prefixing router (based on "github.com/julienschmidt/httprouter") already. Perhaps you could use that one here?

I started out with that one, but I hit some roadblocks... I'll dig back into this and try to remember what those actually _were_ 😉

main.go
hairyhenderson: My thought with `/debug` is that it's not necessarily something you want to expose to the same audience as the other routes. I suppose this is something that can be blocked by reverse proxies, though.

I'll prefix this for now.

main.go
hairyhenderson: Ok, seems reasonable. I'll try that out.

brian-brazil:
> About the -web.external-url: I don't think we need that in Pushgateway because we do not create any absolute URLs containing the host etc.

If we don't need that, do we need web.route-prefix? The prefix was added as an extra configuration option for when you were using an external URL.

beorn7:
@brian-brazil See https://github.com/prometheus/pushgateway/pull/146 for the use case. As the PGW never needs to know about its own hostname in URLs, just the route prefix should do the trick.

main.go
hairyhenderson: > At this point, we have already registered "/-/healthy", "/-/ready", and "/metrics". I think we want them prefixed as well. (At least that's what the Prometheus server is doing.) WDYT?

I've prefixed `/-/*` with `*routePrefix`, but it seems like `*metricsPath` should be unaffected, since it's configured separately.

main.go
hairyhenderson: 🤔 I thought I understood this... I remember going back and forth on this very late at night and I can't remember now why I thought it could cause a problem 🤷‍♂️

hairyhenderson:
@beorn7 I've updated with most of the comments addressed - I'm still looking into `github.com/prometheus/common/route`, and may update this later on if it makes sense.

hairyhenderson:
> I started out with [github.com/prometheus/common/route], but I hit some roadblocks...

Ok, figured out what the roadblocks were. It's mostly around the fact that I'd need to do a bunch of type-casting and wrapping, because `handler.Push()` returns a `httprouter.Handle`, and not a `http.HandlerFunc`. If https://github.com/prometheus/common/blob/master/route/route.go#L42 were exported, it'd be much easier 😉

I'm going to leave this as-is for now, and maybe some day it can be refactored to use `github.com/prometheus/common/route`.

hairyhenderson:
> If https://github.com/prometheus/common/blob/master/route/route.go#L42 were exported, it'd be much easier

I take it back - it's the other way around... not as easy to adapt that way.

hairyhenderson:
All that to say, @beorn7 this is ready to re-review 😉

beorn7:
Thanks. On my review queue (again).

beorn7:
Still on my queue. Sorry for the delay.

hairyhenderson:
@beorn7 no worries - thanks for the update 🙂

main.go
beorn7: Or we could prefix the configured metricsPath, or we can prefix it only if metricsPath is not configured explicitly. But in practice, we should just do what the Prometheus server is doing. Interestingly, at some point it was decided that we don't even need a `--web.telemetry-path` flag anymore for the Prometheus server. But before that flag was removed, the `--web.route-prefix` was used to prefix also the `--web.telemetry-path`. That's one point for doing it in the PGW, too. The other point is that we most likely will eventually remove `--web.telemetry-path` from the PGW, too, In that case, we would definitely need to prefix the (hardcoded) `/metrics` path. Which means we should do it now, too.

Long story short: Please prefix the metrics exposition path, too. Configured or not.

hairyhenderson:
@beorn7 PTAL again - I've addressed the last remaining comment

main.go
```diff
@@ -67,39 +77,45 @@ func main() {
 	// prometheus.EnableCollectChecks(true)
 
 	r := httprouter.New()
-	r.Handler("GET", "/-/healthy", prometheus.InstrumentHandlerFunc("healthy", handler.Healthy(ms)))
-	r.Handler("GET", "/-/ready", prometheus.InstrumentHandlerFunc("ready", handler.Ready(ms)))
+	r.Handler("GET", *routePrefix+"/-/healthy", prometheus.InstrumentHandlerFunc("healthy", handler.Healthy(ms)))
+	r.Handler("GET", *routePrefix+"/-/ready", prometheus.InstrumentHandlerFunc("ready", handler.Ready(ms)))
 
-	r.Handler("GET", *metricsPath, prometheus.Handler())
+	r.Handler("GET", *routePrefix+*metricsPath, prometheus.Handler())
```
beorn7: If `metricsPath` doesn't have a leading `/`, and a `routePrefix` is set, this will result in something weird. Easy way out: Use https://golang.org/pkg/path/#Join

beorn7:
Which has triggered one last follow-up. See above.

main.go
hairyhenderson: 🤦‍♂️ - I _had_ used `path.Join`, but I guess I didn't commit that 😉

hairyhenderson:
Ok, should be good to go now.

beorn7:
Thanks again. Merging now…

hairyhenderson:
Thanks @beorn7 🙂

Is there a 0.4.1 or 0.5.0 in the cards any time soon? I see #143 though it seems there's a dependency on #90 being fixed first...

beorn7:
Yeah, that's a pretty nasty bug I would not like to see in a release again. It's quite high on my priority list, but sadly, there are many items still higher.

PR #150, opened 2018-03-24
Dockerfile updates

### LINKED ISSUE #106: /pushgateway not exists in docker image
/pushgateway not exists in docker image which cause `prom/pushgateway` image run failed using unified container of mesos

```bash
I0308 17:45:46.894984  4941 exec.cpp:162] Version: 1.1.0
I0308 17:45:46.897234  4949 exec.cpp:237] Executor registered on agent f9dbe94e-845c-46aa-a97e-d4aee1fa9eb7-S0
Failed to chdir into current working directory '/pushgateway': No such file or directory
```

hairyhenderson:
Fixes #106 by `mkdir`ing the `/pushgateway` for container runtimes which may not create it the `WORKDIR` automatically.

I've also replaced the [deprecated `MAINTAINER`](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated) instruction with a `LABEL`, to match https://github.com/prometheus/prometheus/blob/master/Dockerfile#L2

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
 FROM        quay.io/prometheus/busybox:latest
-MAINTAINER  The Prometheus Authors <prometheus-developers@googlegroups.com>
+LABEL maintainer "The Prometheus Authors <prometheus-developers@googlegroups.com>"
 
 COPY pushgateway /bin/pushgateway
 
 EXPOSE     9091
+RUN mkdir -p /pushgateway
 WORKDIR    /pushgateway
 ENTRYPOINT [ "/bin/pushgateway" ]


Dockerfile
```diff
@@ -1,8 +1,9 @@
 FROM        quay.io/prometheus/busybox:latest
-MAINTAINER  The Prometheus Authors <prometheus-developers@googlegroups.com>
+LABEL org.opencontainers.image.authors="The Prometheus Authors <prometheus-developers@googlegroups.com>"
```
brian-brazil: Changes like this should be proposed up in the Prometheus repository first and then propagated everywhere else.

hairyhenderson: Thanks. I see in https://github.com/prometheus/prometheus/blob/master/Dockerfile#L2 it's already been changed to:

```docker
LABEL maintainer "The Prometheus Authors <prometheus-developers@googlegroups.com>"
```

I'll change this to match, and suggest using the OCI annotations as a separate issue.

beorn7:
If that fixes things, great.

@discordianfish could you quickly sanity check this? Thanks 1M!

beorn7:
Thanks everybody.

hairyhenderson:
Thanks @beorn7!

PR #151, opened 2018-04-04
Added example for option 'web.listen-address'

maurorappa:
it's not intuitive you need to use the format "IP:PORT", for example this invocation:

`docker run -ti prom/pushgateway -web.listen-address 8080`

produce a misleading error:

`INFO[0000] Starting pushgateway (version=0.4.0, branch=master, revision=6ceb4a19fa85ac2d6c2d386c144566fb1ede1f6c)  source=main.go:57
INFO[0000] Build context (go=go1.8.3, user=root@87741d1b66a9, date=20170609-12:25:40)  source=main.go:58
INFO[0000] Listening on 8080.                            source=main.go:102
FATA[0000] listen tcp: address 8080: missing port in address  source=main.go:105`

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ If you want to compile yourself from the sources, you need a working Go
 setup. Then use the provided Makefile (type `make`).
 
 For the most basic setup, just start the binary. To change the address
-to listen on, use the `-web.listen-address` flag. By default, Pushgateway 
+to listen on, use the `-web.listen-address` flag (ie "0.0.0.0:9090"). By default, Pushgateway
 does not persist metrics. However, the `-persistence.file` flag
 allows you to specify a file in which the pushed metrics will be
 persisted (so that they survive restarts of the Pushgateway).
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ If you want to compile yourself from the sources, you need a working Go
 setup. Then use the provided Makefile (type `make`).
 
 For the most basic setup, just start the binary. To change the address
-to listen on, use the `-web.listen-address` flag (i.e. "0.0.0.0:9091" or ":9091").
+to listen on, use the `-web.listen-address` flag (e.g. "0.0.0.0:9091" or ":9091").
 By default, Pushgateway does not persist metrics. However, the `-persistence.file` flag
 allows you to specify a file in which the pushed metrics will be
 persisted (so that they survive restarts of the Pushgateway).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ If you want to compile yourself from the sources, you need a working Go
 setup. Then use the provided Makefile (type `make`).
 
 For the most basic setup, just start the binary. To change the address
-to listen on, use the `-web.listen-address` flag. By default, Pushgateway 
+to listen on, use the `-web.listen-address` flag (ie "0.0.0.0:9090"). By default, Pushgateway
```
hairyhenderson: Not sure `9090` is the best port to have in the README - that's most commonly used by Prometheus. How about using the actual default in the example:

```patch
+to listen on, use the `-web.listen-address` flag (e.g. "0.0.0.0:9091" or ":9091"). By default, Pushgateway
```

maurorappa: True, gimme a port which is not the standard one ;)

hairyhenderson:
LGTM, FWIW 😉

README.md
```diff
@@ -40,7 +40,7 @@ If you want to compile yourself from the sources, you need a working Go
 setup. Then use the provided Makefile (type `make`).
 
 For the most basic setup, just start the binary. To change the address
-to listen on, use the `-web.listen-address` flag. By default, Pushgateway 
+to listen on, use the `-web.listen-address` flag (ie "0.0.0.0:9091" or ":9091"). By default, Pushgateway
```
beorn7: Nit: Please write "i.e." and re-wrap the lines to 80 chars for consistency with the rest of the file.

maurorappa:
@beorn7, I amened,  the reason I raised to this PR was the fact the error is measleading if you don't use the right (IP:PORT) format:
"FATA[0000] listen tcp: address 8080: missing port in address"

now I could have raised a PR to change the code or change the Readme :)

README.md
brian-brazil: Isn't this an e.g. rather than an i.e.? You can use values other than these.

beorn7: Oops, yeah, @brian-brazil is right, this should be an "e.g.". And sure, other examples than the default might make more sense, but I don't feel strongly…

beorn7:
That error message is coming from the Go standard library, so you would need to change their code.
(But given that this is heavily used in exactly that way in many places, perhaps the error message is not that misleading after all.)

README.md
beorn7: To clarify: @brian-brazil wasn't suggesting to actually use other values but only tried to explain why it is "e.g." rather than "i.e.".

So @maurorappa , my apologies for the 2nd-order nit, but if you could just change "i.e." to "e.g.", we are good to go.

beorn7:
Thanks! And sorry again for the nit-picky round-trips.

PR #153, opened 2018-04-19
 Use X-Forwarded-For header for instance, if present

rayslinky:


brian-brazil:
Why do you want this? You should probably read https://prometheus.io/docs/practices/pushing/, things pushed to the pushgateway should not have instance labels.

beorn7:
Yeah, we really discourage the usage of an instance label.

Setting an instance label automatically is, indeed, considered legacy behavior. And that's why this functionality is in `LegacyPush`, a function that is only here to reconstruct legacy behavior for backwards compatibility. It would defeat that purpose if we changed `LegacyPush`.

Issue #155, opened 2018-04-28
Failed test: diskmetricstore_test.go when running "make"


LittleFatz:
The detail of the issue is as below:
[root@6641fb4fae48 pushgateway]# make
>> formatting code
>> building binaries
 >   pushgateway
>> running tests
?       github.com/prometheus/pushgateway       [no test files]
ok      github.com/prometheus/pushgateway/handler       0.006s
--- FAIL: TestAddDeletePersistRestore (0.17s)
        diskmetricstore_test.go:435: Expected timestamp 2018-04-27 10:07:25.122420715 +0000 UTC m=+2.002244876, got 2018-04-27 10:07:25.122420715 +00
0 UTC.
FAIL
FAIL    github.com/prometheus/pushgateway/storage       0.201s
make: *** [test] Error 1

PR #157, opened 2018-04-28
Fix time comparison in test

juliusv:
time.Time as obtained by time.Now() contains a monotonic clock value
that we don't persist, so the original and stored time will not be equal
under "==" / "!=" comparison operators. Using Equal() handles missing
monotonic clock fields on otherwise identical time.Time values
correctly.

Fixes https://github.com/prometheus/pushgateway/issues/155


beorn7:
:+1:

PR #167, opened 2018-05-19
WIP: Update vendoring

SuperQ:
* Update `github.com/prometheus/...`.
* Update `golang.org/x/sys/...`.

Closes: https://github.com/prometheus/pushgateway/issues/166

Signed-off-by: Ben Kochie <superq@gmail.com>

PR #168, opened 2018-05-19
Make it work with current client_golang

### LINKED ISSUE #163: Don't log persisting if persistence is disabled.
If no persistence file is provided, no persisting is happening, see https://github.com/prometheus/pushgateway/blob/master/storage/diskmetricstore.go#L305

However, it is still logged: https://github.com/prometheus/pushgateway/blob/master/storage/diskmetricstore.go#L176

### LINKED ISSUE #166: Uses deprecated client_golang API
Hi,

The pushgateway has an old version of the client_golang vendored, and uses deprecated API methods that have now been removed from the library:

```
src/github.com/prometheus/pushgateway/main.go:67:2: undefined: prometheus.SetMetricFamilyInjectionHook
```

Please, can you update the vendoring and replace that call with the current API?

beorn7:
@SuperQ @TheTincho This also fixes a bunch of other trivial things. I hope you don't mind reviewing them, too. The commits are all nicely separate.

Fixes #163 .
Fixes #166 

As noted in the final commit, this does not yet remove the "soft deprecated" parts of client_golang. That's a bit more involved as it requires thinking of how to change the HTTP metrics. But those things are still supported by the current client_golang (and will be by the upcoming v0.9 of it).


beorn7:
I think this should supercede #167 by now (all needed to fix the CI...)

PR #178, opened 2018-05-25
Removed forward slash from delete ajax call url.

### LINKED ISSUE #177: Delete Group button not working
Delete Group Button is not working with `web.route-prefix`, on clicking it's raising an alert saying `Deleting metric group failed: Forbidden`.

tempsnip

This is happening due to https://github.com/prometheus/pushgateway/blob/master/resources/static/functions.js#L52, there is no `web.route-prefix` prepend to ajax call url.

manthansharma:
Fixed #177


beorn7:
Thanks. :+1:

PR #180, opened 2018-05-29
Fix persistence conversion bug

beorn7:
Fixes #179 .

@SuperQ @dneuhaeuser-zalando

Issue #176, opened 2018-05-24
Useless/confusing flags displayed on status page


TheTincho:
Hi,

With the move to kingpin, the status page now is showing the values for flags that only make sense as one-offs in the command line, and can be confusing for users, like 'help', 'version', 'completion-bash', and a few others.

brian-brazil:
We filtered these on Prometheus, so could do the same here.

PR #181, opened 2018-05-29
Update client_golang/prometheus vendoring for bugfix

beorn7:
Signed-off-by: beorn7 <beorn@soundcloud.com>

@SuperQ step 2 as promised in https://github.com/prometheus/pushgateway/pull/180

After that, we should be able to cut 0.5,1


beorn7:
Wait, I'll also fix #177 and #176

PR #182, opened 2018-05-29
Fix flag display

### LINKED ISSUE #176: Useless/confusing flags displayed on status page

beorn7:
@SuperQ as promised in person.

@TheTincho fixes #176 .

Web asset regeneration not strictly needed for this, but it wasn't done for #178 .

### DIFF
--- a/bindata.go
+++ b/bindata.go
@@ -95,7 +95,7 @@ func staticBootstrap334DistLicense() (*asset, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


TheTincho:
Thanks!

PR #198, opened 2018-08-27
Migrate from go-bindata to vfsgen

### LINKED ISSUE #193: Replace go-bindata
See https://github.com/prometheus/prometheus/issues/2411#issuecomment-404445255 and https://github.com/prometheus/prometheus/pull/4430

TL;DR `go-bindata` isn't maintained anymore and needs to be replaced by an equivalent project in better shape.

simonpasquier:
Closes #193 

The PR is based on https://github.com/prometheus/prometheus/pull/4430.


beorn7:
Great, thanks.

BTW: I still see a diff in file timestamps if I run `make assets`. I thought vfsgen would avoid that.

But anyway, nothing gets worse in that regard.

simonpasquier:
> BTW: I still see a diff in file timestamps if I run make assets. I thought vfsgen would avoid that.

Hmm, I don't see the same thing.

beorn7:
Most likely because your files still have the same timestamp an when you ran `make assets` for this PR. My files have different timestamps, and thus if I run `make assets`, I get a timestamp diff (despite the content of the files being equal).

simonpasquier:
Ah you're right, I've been fooled by my local filesystem! https://github.com/shurcooL/vfsgen/issues/26 is the upstream issue. I might experiment with something.

PR #204, opened 2018-09-14
Update vendoring and adjust to new client_golang

beorn7:
- Update all the vendored dependencies to current state.
- Then remove all uses of now deprecated types from client_golang.
- Document the exposed metrics (as they have changed significantly due to the above).

@stuartnelson3 @SuperQ This has two purposes:
- Pull in new client_golang behavior, in particular the relaxed check for label consistency.
- Try out the current state of client_golang before releasing it. (I plan the same for node_exporter and prometheus itself to cover other use cases.)


SuperQ:
What do you think about moving the 3rd party metrics to a different metrics path like `/pushgateway`. Similar to how we have separate paths for blackbox `/probe`.

This would avoid the conflicts between the pushgateway's internal metrics and metrics that a client wants to push.

beorn7:
That's something we might consider. but it would be a fairly fundamental change. surely not in scope for this PR.
Also, most of the time people run into this, they try to turn Prometheus into a push-based system. Thus, I'm not even sure if that's desirable to facilitate.
The people at SC use our many Pushgateways in the way it is meant to be, and they never ran into collisions with the PGW's own metrics.

Issue #199, opened 2018-08-29
MAINTAINER instruction in Dockerfile is deprecated


tariq1890:
It should be in the following format
` LABEL maintainer="..."`

See https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

PR #212, opened 2018-10-05
Improves Dockerfile format

ChrsMark:
closes https://github.com/prometheus/pushgateway/issues/199

- Changing LABEL instruction according to https://docs.docker.com/engine/reference/builder/#maintainer-deprecated
- Formating

cc: @beorn7


beorn7:
My apologies. This fell of my radar without noticing.
:+1:

PR #218, opened 2018-10-18
Fix flaky test

beorn7:
As the pick of the help string is random, we have to be a bit more involved with the test.

### DIFF
--- a/storage/diskmetricstore_test.go
+++ b/storage/diskmetricstore_test.go
@@ -748,10 +748,33 @@ func TestHelpStringFix(t *testing.T) {
 	time.Sleep(20 * time.Millisecond) // Give loop() time to process.
 
 	// Either we have settle on the mfh1 help string or the mfh2 help string.
-	err1 := checkMetricFamilies(dms, mfggFixed, mfh21)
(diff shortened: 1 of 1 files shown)


storage/diskmetricstore_test.go
```diff
@@ -783,7 +806,7 @@ func checkMetricFamilies(dms *DiskMetricStore, expectedMFs ...*dto.MetricFamily)
 	for i, got := range gotMFsAsStrings {
 		expected := expectedMFsAsStrings[i]
 		if expected != got {
-			return fmt.Errorf("expected metric family %#v, got %#v", expected, got)
+			return fmt.Errorf("expected metric family '%s', got '%s'", expected, got)
```
grobie:
```suggestion
			return fmt.Errorf("expected metric family %q, got %q", expected, got)
```

beorn7: I know it's tempting to suggest that, but the `'%s'` is intentional as the double-quoted metric families look horrible. Their string representation contains a lot of double quotes on their own, so `%q` escapes them all. It looks like 30% backslashes in the error message...

beorn7: A `%q` that uses single quotes would be cool.

PR #219, opened 2018-11-05
Fix Dockerfile to be able to build and release without golang installed

Naouak:
Hopefully, it will fix the build on Docker hub.


beorn7:
Thanks. However, I'd like to keep the Dockerfile here in sync with what we do in the many other repos in the Prometheus org.

What problem are you trying to fix here exactly? Is this a problem that all the other repos have, too?

Naouak:
I'm trying to fix the broken build on docker hub and provide an easy way to build it for people wanting to use it in their project in a non go stack.

beorn7:
Is the build on docker hub broken in the same way for all the other Prometheus repos (e.g. prometheus/prometheus or prometheus/alertmanager)?

brian-brazil:
Yes, we stopped using it about 2 years ago when we switched to promu and doing our own Docker builds. Only one of the Java repos uses it now.

beorn7:
What part of the documentation are you referring to?

I can only find https://github.com/prometheus/pushgateway#using-docker , and that's about _running_ PGW, not about _building_ it. The instructions given there work perfectly fine for me to run the latest released version of the PGW.

Naouak:
The dockerhub link goes to prom/pushgateway which has failing build for 2
years.

beorn7:
That's because we upload the readily built artifacts and don't use Docker build. What you get if you run `docker pull prom/pushgateway` is the most current PGW release.

beorn7:
The is unfortunately no way to hide the "Build" tag. But the automatic builds are deactivated, that's why no build was even attempted in the last 2 years, despite releases and updates to the container happening.

Naouak:
Have you tried to contact Docker inc. about that? This should be considered
as a bug IMO from their part that any new image is not referenced on the
build page as it is highly confusing.

sdurrheimer:
It was a autonomated build, but the docker hub doesn't let us transform a
automated build repository into a normal repository.

The only way is to delete the repository and recreate it as a normal
repository but would need to re-import all the tag history.

PR #220, opened 2018-11-16
circleci: switch to 2.1 config

simonpasquier:
Similar to https://github.com/prometheus/prometheus/pull/4713 and https://github.com/prometheus/alertmanager/pull/1579. Once the 2.1 config is enabled, we can reduce duplication in `.circleci/config.yml` a bit.


beorn7:
Cool. Thanks.

PR #221, opened 2018-11-16
*: support Go modules

simonpasquier:
Updating to the latest `Makefile.common` implies to modify the CircleCI configuration that pushes the container images (but in the end, the YAML is less repetitive).


simonpasquier:
Yep, I've noticed that too but didn't find any concrete information about it. I guess I'll have to dig the Go issues and source code to find the answer...

PR #226, opened 2018-12-04
Improve documentation WRT grouping

### LINKED ISSUE #225: Clarify documentation that DELETE only ever deletes by grouping key
I think it's regrettable you cannot remove all metrics by only specifying the job.
Multiple people have run into this: (e.g. https://github.com/prometheus/pushgateway/issues/128, https://github.com/prometheus/pushgateway/issues/162).
The confusion is only increased because the documentation is not correct:
It still mentions in the README.md you _can_ delete metrics by job.

Not having this feature is annoying because sometimes you want to clear the old metrics without knowing what the grouping keys were.

beorn7:
Also, delete the deprecated parts, as they have been declared so long ago.

@praetp

Fixed #225 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,36 +215,6 @@ even if escaped as `%2F`. (The decoding happens before the path
 routing kicks in, cf. the Go documentation of
 [`URL.Path`](http://golang.org/pkg/net/url/#URL).)
 
-### Deprecated URL
-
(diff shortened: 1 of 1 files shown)


beorn7:
The actual support for the legacy HTTP API is removed in https://github.com/prometheus/pushgateway/pull/227

PR #234, opened 2019-02-04
code and test coverage improvements for diskmetricstore

tariq1890:


storage/diskmetricstore_test.go
```diff
@@ -812,6 +873,27 @@ func checkMetricFamilies(dms *DiskMetricStore, expectedMFs ...*dto.MetricFamily)
...
+func checkMetricFamilyGroups(dms *DiskMetricStore, expectedMFMap map[uint64]map[string]string) error {
+	mfMap := dms.GetMetricFamiliesMap()
+
+	if expected, got := len(expectedMFMap), len(mfMap); expected != got {
+		return fmt.Errorf("expected %d metric families in map, but got %d", expected, got)
+	}
+
+	for k, v := range mfMap {
+		if innerMap, ok := expectedMFMap[k]; ok {
+			for metricName, metricString := range innerMap {
```
beorn7: You should probably also assess that `len(innerMap) == len(v.Metrics)`. Otherwise, you would miss if `v.Metrics` contains more metrics than expected.

tariq1890: True . Thanks for pointing that out :)

tariq1890: This has been done.

PR #235, opened 2019-02-09
Add command-line option to clear all metrics after scrape requests

berendo:
Having read mailing list discussions (as well as the `README.md`) on various pros/cons of TTL-like features, I have implemented a command-line option for a _non-TTL-based_ immediate clearing of all existing metrics accumulated in the push gateway right after every scrape request.

We needed this for our own purposes and if the maintainers find that it falls on the non-controversial side of the TTL-related discussion, please feel free to merge it. Philosophically, this implementation is not _that_ different than using the `textfile` collector and blanking the text file every time a `GET /metrics` request is observed via a potential proxy sitting in front of `node_exporter`, so I don't think it's enabling a usage that's not otherwise possible by combining a few components...

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,6 +50,7 @@ func main() {
 		listenAddress       = app.Flag("web.listen-address", "Address to listen on for the web interface, API, and telemetry.").Default(":9091").String()
 		metricsPath         = app.Flag("web.telemetry-path", "Path under which to expose metrics.").Default("/metrics").String()
 		routePrefix         = app.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints.").Default("").String()
+		deleteAfterScrape   = app.Flag("web.delete-after-scrape", "").Bool()
 		persistenceFile     = app.Flag("persistence.file", "File to persist metrics. If empty, metrics are only kept in memory.").Default("").String()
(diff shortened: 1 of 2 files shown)


beorn7:
Thanks for your efforts. Unfortunately, your idea is directly at odds with a central Prometheus concept, namely that you can add scrapers at will without interfering with the metric collection of other scrapers. For example, you want to test some rules on your laptop, just spin up a Prometheus server there and let it scrape your production instances. Or think about the HA concept of Prometheus: Run _n_ servers with the same config in parallel, all scraping the same targets. Deleting metrics after each scrape would totally upset metrics collection for all other scrapers.

Sorry to disappoint you here.

berendo:
> For example, you want to test some rules on your laptop, just spin up a Prometheus server there and let it scrape your production instances. Or think about the HA concept of Prometheus:

Absolutely. I'm aware of the implications and appreciate Prometheus developers' desire to protect users from self-harm, hence my personal implementation as a command-line option. I expect to use this on only certain types of nodes which can tolerate metrics loss in the manner you're alluding to and mostly submitted the PR as a "just in case." Thanks for your consideration.

PR #236, opened 2019-02-10
Add command-line option to track scrape requesters for each metric

berendo:
This is a follow-up to https://github.com/prometheus/pushgateway/pull/235 and while I understand that @beorn7 may indeed make the same final judgment, his comments did make me alter my implementation a bit to track the scrape requesters for each metric separately and implement the "one shot" propagation of every metric on a per-requester (at the IP address level) basis. This allows the `pushgateway`, when launched with the `--web.scrape-once` option, to "hide" each metric from a given scraper that has already scraped it _until_ that metric is pushed again. This addresses the concerns raised in https://github.com/prometheus/pushgateway/pull/235 related to HA-like scenarios whereby it is desirable for the presence of any number of scrapers to not affect the propagation of metrics to other scrapers. It also leaves the on-disk persistence behavior of `pushgateway` unaltered.

tl;dr @beorn7 thank you for your earlier feedback and I won't be offended if you close this PR as well, but it seemed amiss not to submit the revision at this point.

### DIFF
--- a/handler/handler_test.go
+++ b/handler/handler_test.go
@@ -35,7 +35,7 @@ func (m *MockMetricStore) SubmitWriteRequest(req storage.WriteRequest) {
 	m.lastWriteRequest = req
 }
 
-func (m *MockMetricStore) GetMetricFamilies() []*dto.MetricFamily {
+func (m *MockMetricStore) GetMetricFamilies(querier string) []*dto.MetricFamily {
 	panic("not implemented")
 }
 
--- a/main.go
+++ b/main.go
@@ -50,6 +50,7 @@ func main() {
 		listenAddress       = app.Flag("web.listen-address", "Address to listen on for the web interface, API, and telemetry.").Default(":9091").String()
 		metricsPath         = app.Flag("web.telemetry-path", "Path under which to expose metrics.").Default("/metrics").String()
 		routePrefix         = app.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints.").Default("").String()
+		scrapeOnce          = app.Flag("web.scrape-once", "Allow each metric to be scraped only once by every scraper until the metric is pushed again.").Bool()
 		persistenceFile     = app.Flag("persistence.file", "File to persist metrics. If empty, metrics are only kept in memory.").Default("").String()
 		persistenceInterval = app.Flag("persistence.interval", "The minimum interval at which to write out the persistence file.").Default("5m").Duration()
 	)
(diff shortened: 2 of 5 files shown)


beorn7:
Yeah, I'm sorry. Even if you are patching up one failure mode, this is still a fundamental violation of one of the most fundamental concepts of Prometheus. I could go on with concrete examples, like what happens if the scraping server decides to double the scrape frequency? Suddenly metrics disappear every other scrape. Which assumes in the first place that push and scrape frequencies are somehow in sync, which is by itself already a brittle setup. What happens if scrapers come in through a reverse proxy (e.g. to add TLS) and now all appear to have the same IP? But the examples only serve the purpose of why that fundamental concept of Prometheus is in place. Even if you create a work-around for every single one of them, you are still asking to change a low-numbered article in Prometheus's constitution. We wouldn't be able to do that in a PR anyway, so you would need to go to the [developers mailing list](https://groups.google.com/forum/#!forum/prometheus-developers) and convince the community there. (Hint: _Very_ unlikely to succeed. And that applies to everybody, not just Brian…)

Thus, I'm closing this. If you want to discuss the topic further, the [developers mailing list](https://groups.google.com/forum/#!forum/prometheus-developers) is the place to do so.

berendo:
Understood @beorn7, thank you again. I realized this was “controversial,” but wanted to forward to work anyway. For posterity’s sake, let me answer two questions you asked:

> like what happens if the scraping server decides to double the scrape frequency? Suddenly metrics disappear every other scrape. Which assumes in the first place that push and scrape frequencies are somehow in sync,

I’m not sure I see that. With my implementation, there’s no assumption of push vs scrape frequencies, just that each scraper can observe a _stable_ value once and only once. A given observers frequency has no impact on the observation of other scrapers, and I’m assuming you’ll agree that even with a “pure” Prometheus setup the selection of the scrape frequency has _some_ effect on the observed data.

> What happens if scrapers come in through a reverse proxy (e.g. to add TLS) and now all appear to have the same IP?

Yeah, I considered this, and I even looked at the scraper-side code to see if there are any `X-Prometheus-*` headers that provide a UUID or similar identifying information on a per scraper basis. I also thought about your HA scenario and thought that having two or more scrapers that then need to share a single point of failure may not be that HA, so I compromised and used the IP address. It’s not idel, I agree.

beorn7:
> each scraper can observe a _stable_ value once and only once

That "once and only once" is probably as direct as a contradiction with the Prometheus model can go.
Prometheus is not for event logging.

berendo:
> Prometheus is not for event logging.

Got it. That’s not actually my intent, as I’m collecting strictly numeric metrics, but I can see how “one shot”/“impulse” metrics can be seen as “event logging.” Thanks again for the helpful discussion.

PR #239, opened 2019-03-27
Fix the Route Prefix bug and add ExternalURL Support (#190)

### LINKED ISSUE #190: Add/modify route-prefix and external-url flags to work the same as for the Prometheus server
Details are discussed in #149 . This issue is meant to track the (more urgent) user-facing behavior (in contrast to implementation details of the routing).

In short: `--web.route-prefix` should work in the same way as for the Prometheus server. `--web.external-url` should be added with the same semantics as for the Prometheus server.

apghero:
This fixes #190, I believe.

I've gotta figure out how to add in the `--web.external-url` with the same semantics as prometheus still. But, this should address the route-prefix issue as it stands right now.

Happy to address the externalURL in another PR entirely, in which case this wouldn't be work in progress anymore, and would be ready for review.


apghero:
@beorn7: This should be ready to take a look at when you have a chance.

I did notice that the asset compilation doesn't work for repositories outside of the `$GOPATH`, which, since this repo is using modules shouldn't be required. I'll file another bug for that.

beorn7:
Many thanks, I'll have a look ASAP.

main.go
```diff
@@ -135,6 +135,25 @@ func handlePprof(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
 	}
 }
 
+func computeRoutePrefix(prefix string, externalURL *url.URL) string {
```
beorn7: The inline comments in this function are essentially just describing the (very simple) code and thus make it actually less readable. I guess the point you are trying to make is that this code implements the spec given by the flags. How about turning them into a doc comment before the function:

```Go
// computeRoutePrefix returns the effective route prefix based on the provided
// flag values for --web.route-prefix and--web.external-url. With prefix empty,
// the path of externalURL is used instead. A prefix "/" results in an empty
// returned prefix. Any non-empty prefix is normalized to start, but not to end,
// with "/".
```

handler/status.go
```diff
@@ -65,6 +67,14 @@ func Status(
 				"value": func(f float64) string {
 					return strconv.FormatFloat(f, 'f', -1, 64)
 				},
+				"join": func(base string, paths ...string) string {
```
beorn7: Instead of passing in a whole new function, wouldn't it be easier to just pass in the pre-calculated complete string to prefix URLs in the template with? The function is always called in the same way in the template.

apghero:
Oh boy. Some of these things are silly (copyright notice). I'll fix 'em up.

handler/status.go
apghero: I can do that. The general purpose function is useful for future uses, but those uses don't currently exist, and are unlikely to be drastically different I'm guessing.

apghero:
OK, BaseURL is precomputed before template render.

beorn7:
Many thanks. Let's hope this _really_ fixes the issue for good.

nodox:
Have we stopped trying to release this fix? Would love this functionality but it seems the latest release was in December 2018.

beorn7:
A releases will happen soon. For this particular feature, I would really like to see it being tried out a bit by early adopters, who build from the tip of master. @nodox would be great if you volunteered.

nodox:
Building from master might be too involved from me. However if you create an docker image tagged with the changes in master you've got yourself a beta tester. Right now I'm trying to make this work using the Prometheus Helm chart on a Kubernetes. @beorn7

sepich:
Thank you,
at least now I've got it working with image `prom/pushgateway:master`,
args: `--web.route-prefix=/push/`,
and nginx location:
```
location /push {
    proxy_pass http://pushgateway:9091;
}
```
REST API also successfully receiving POSTs on `host.domain.local/push/`

When i've tried with `--web.route-prefix=https://host.domain.local/push/` (same as for prometheus) - then all links to static resources became: `/host.domain.local/push/script.js`

beorn7:
Thanks @sepich . I'll have a look ASAP.

beorn7:
For me, both Prometheus and PGW panic if I set `--web.route-prefix=https://host.domain.local/push/`:
```
panic: wildcards must be named with a non-empty name in path '/https://host.domain.local/push/-/healthy'
```

I think rejecting it is the right thing as `--web.route-prefix` is just a path prefix, not a whole URL. We could make the error message nicer, of course.

@sepich did you manage to avoid the panic somehow?

apghero:
@sepich have you tried `--web.external-url=https://host.domain.local/push/` ? That should do the right thing as it takes the path of the external-url and puts it as the `--web.route-prefix`. 

I agree with @beorn7 that the prefix should be better validated.

sepich:
Ups, sorry for the typo.
I've meant `--web.external-url=https://host.domain.local/push/` leads to:
`<link rel="stylesheet" href="https:/host.domain.local/push/static/bootstrap-3.3.4-dist/css/bootstrap.min.css">` 
note single slash after `https:`,
which Chrome interpret as:
`GET https://host.domain.local/host.domain.local/push/static/bootstrap-3.3.4-dist/css/bootstrap.min.css net::ERR_ABORTED 404`

Yes, the `--web.route-prefix=/push/` works fine.

beorn7:
I see. I'll try and fix.

beorn7:
#244 should fix it.

PR #242, opened 2019-04-11
Run process as non-root user

tfluehmann:
A process should not run as root in a container if not explicitly required. Therefore, added a pushgateway user with user id 1000 and set
the user to 1000. This should be compatible with k8s PSP's which check
for the user id.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,12 @@
 FROM quay.io/prometheus/busybox:latest
 LABEL maintainer="The Prometheus Authors <prometheus-developers@googlegroups.com>"
 
-COPY pushgateway /bin/pushgateway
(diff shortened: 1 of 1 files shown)


beorn7:
Thanks for catching this. AFAIK this problem was already fixed for `prometheus/node_exporter/Dockerfile` and `prometheus/prometheus/Dockerfile`. However, in those cases, the user was changed from `root` to `nobody`. Do you think that would be possible here, too? I would like to keep things consistent with other Prometheus repos.

tfluehmann:
Hi beorn7 
Thanks for the quick feedback. I didn't knew that so far. Yes of course, I'll change it to nobody.

beorn7:
Thanks.

Issue #240, opened 2019-03-28
`make assets` fails when repository not in $GOPATH


apghero:
## Bug Report
**What did you do?**

When attempting to recompile assets (`make assets`) it failed to work with my repository outside of the `$GOPATH`

**What did you expect to see?**

I expected to see assets compiled successfully.

**What did you see instead? Under which circumstances?**

```
 make assets                                                                                                           
>> writing assets                                                                                                                                               
2019/03/27 17:29:53 cannot find package "github.com/prometheus/pushgateway/resources" in any of:                                                                
        /usr/local/go/src/github.com/prometheus/pushgateway/resources (from $GOROOT)                                                                            
        /Users/agwozdziewycz/go/src/github.com/prometheus/pushgateway/resources (from $GOPATH)                                                                  
exit status 1                                                                                                                                                   
doc.go:22: running "go": exit status 1                                                                                                                          
make: *** [assets] Error 1                                                          
```

**Environment**

```
$ env | grep GO
GOPATH=/Users/agwozdziewycz/go
```

Nothing else should be relevant.

* System information:

		Darwin 18.2.0 x86_64

* Pushgateway version:

```
$ ./pushgateway --version
pushgateway, version 0.7.0 (branch: apghero/bugfix/190/route-prefix-bug, revision: 153da9be52b7851c67e552f7635ab6c91d209a60)
  build user:       agwozdziewycz@agwozdziew-ltm.internal.salesforce.com
  build date:       20190328-03:34:09
  go version:       go1.12.1
```

* Pushgateway command line:

		N/A.

apghero:
Note that there is a workaround for this:

```
$ ln -s /Users/agwozdziewycz/Devel/pushgateway/resources/ ~/go/src/github.com/prometheus/pushgateway/resources
```

beorn7:
I just tried to reproduce the problem in the following way:

```
$ mkdir not_in_GOPATH
$ cd not_in_GOPATH/
$ git clone git@github.com:prometheus/pushgateway
Cloning into 'pushgateway'...
remote: Enumerating objects: 24, done.
remote: Counting objects: 100% (24/24), done.
remote: Compressing objects: 100% (24/24), done.
remote: Total 2474 (delta 4), reused 14 (delta 0), pack-reused 2450
Receiving objects: 100% (2474/2474), 3.51 MiB | 6.52 MiB/s, done.
Resolving deltas: 100% (1392/1392), done.
$ cd pushgateway/
$ make assets
[...]

PR #243, opened 2019-04-11
Updates to Makefiles

beorn7:
@apghero does this perhaps help you with #240 ?


beorn7:
OK, so that's not enough as the `go run` called from within `go generate` is misbehaving if a `GOPATH` is set.

beorn7:
Last commit added here should really fix #240. @simonpasquier I hope that makes sense. Not sure why we needed the more complicated path calculation in the first place. Perhaps you know something about it.

apghero:
@beorn7 this works without the symlink in my GOPATH. :+1:

beorn7:
OK, merging now. @simonpasquier please let me know if you have any concerns.

simonpasquier:
@beorn7 all good for me!

> Not sure why we needed the more complicated path calculation in the first place. Perhaps you know something about it.

IIRC it was taken from the https://github.com/shurcooL/vfsgen examples but indeed we simplified it the same way in [prometheus/prometheus](https://github.com/prometheus/prometheus/commit/d89c01b0658465523cca1986a584e2598488e315).

PR #244, opened 2019-04-11
Replace `path.Join` by normal string concatenation

beorn7:
`path.Join` cleans up all double-slashes anh thus wreaks havoc with a
URL. The plain string concatenation is actually fine here as all the
sanitations have already happened.

This is a fix for the [fix](#239) of #190.

@apghero @sepich


apghero:
That looks fine to me.

beorn7:
Thanks. Let's merge…

PR #245, opened 2019-04-12
Cut v0.8.0

beorn7:
Signed-off-by: Bjoern Rabenstein <bjoern@rabenste.in>


beorn7:
@apghero @tfluehmann @sepich @nodox FYI

@simonpasquier last chance to intervene before releasing #243 .

beorn7:
Also added upgrade to Go1.12

sepich:
Re-tested `--web.external-url=https://host.domain.local/push/` on (version=0.7.0, branch=master, revision=f355ade9d6b708c7f2e7b6c04bdb4ceea0201671).
Works fine, thank you.

Only note is that previous persisted data was not loaded, because now file should be owned by nobody:
`time="2019-04-12T19:28:57Z" level=error msg="Could not load persisted metrics: open /data/pushgateway.dat: permission denied" source="diskmetricstore.go:82"`
Maybe this should be mentioned in Release notes.

beorn7:
Very good point on the Docker issue. I'll mark the user change as a `[CHANGE]` then and will point it out in a separate note.

WRT the copyright notice: There is just no point in updating this (or others). Our policy so far is to use the current year in newly created files, and leave it where it is in existing files. The legal impact is anyway disputed at best.

beorn7:
Reshuffled commits and added note about Dockerfile change.

SuperQ:
I'd like to add a vendoring bump first.

beorn7:
I'd prefer not to do a blind vendoring bump just before a release. Unless, of course, there is something specific we need, like known bugs. I did a quick check for that a few days ago, and nothing stuck out.

A "update everything because we can" might make sense now and then to not fall behind too much, but it has some risk, and thus I would not like to do it just before a release. Note that Go modules are deliberately designed to avoid the "let's use the newest of everything" pattern.

SuperQ:
Alright, up to you.

PR #246, opened 2019-04-13
Update vendoring

SuperQ:
* Bump Go to 1.12.
* Update all vendoring.


beorn7:
Let's not do this just before a release.

PR #252, opened 2019-05-09
Use common/server/StaticFileServer

beorn7:
@brian-brazil @mrasu same as https://github.com/prometheus/prometheus/pull/5551 

Also, updates the dependencies in general, which creates the many changed files.


brian-brazil:
:+1:

PR #263, opened 2019-06-26
Use promlog to support --log.format=json flag

kfdm:
Similar to https://github.com/prometheus/alertmanager/pull/1658, updates pushgateway to support json logging.

I'm having minor problem with the last line so any assistance is appreciated :) 

```go
level.Info(logger).Log("msg", "Received SIGINT/SIGTERM; exiting gracefully...")
```


kfdm:
I realize that many of the other logging statements will probably need to be updated as well. It looks like Alertmanager passes the logger component around so I would appreciate guidance on how you would like me to handle that here 🙇

main.go
```diff
@@ -160,6 +165,6 @@ func interruptHandler(l net.Listener) {
 	notifier := make(chan os.Signal, 1)
 	signal.Notify(notifier, os.Interrupt, syscall.SIGTERM)
 	<-notifier
-	log.Info("Received SIGINT/SIGTERM; exiting gracefully...")
+	// level.Info(logger).Log("msg", "Received SIGINT/SIGTERM; exiting gracefully...")
```
beorn7: Why is this commented now?

main.go
```diff
@@ -54,18 +57,20 @@ func main() {
...
 	*routePrefix = computeRoutePrefix(*routePrefix, *externalURL)
 
-	log.Infoln("Starting pushgateway", version.Info())
-	log.Infoln("Build context", version.BuildContext())
-	log.Debugf("Prefix path is '%s'", *routePrefix)
-	log.Debugf("External URL is '%s'", *externalURL)
+	level.Info(logger).Log("msg", "Starting pushgateway", "version", version.Info())
+	level.Info(logger).Log("build_context", version.BuildContext())
+	level.Debug(logger).Log("msg", "Prefix path", "path", *routePrefix)
+	level.Debug(logger).Log("msg", "Extenral URL", "url", *externalURL)
```
beorn7: Typo: Extenral

main.go
```diff
@@ -105,20 +110,20 @@ func main() {
 	// Re-enable pprof.
 	r.GET(*routePrefix+"/debug/pprof/*pprof", handlePprof)
 
-	log.Infof("Listening on %s.", *listenAddress)
+	level.Info(logger).Log("listen_address", *listenAddress)
 	l, err := net.Listen("tcp", *listenAddress)
 	if err != nil {
 		log.Fatal(err)
```
beorn7: This one is still using the old logging.

beorn7:
Sorry, only now I have seen the comments you already added. Will reply in a sec.

main.go
kfdm: This is commented out because I need to figure out how to pass the `logger` instance in there.

main.go
kfdm: Yes ^^;

main.go
kfdm: Yeah, it looks like internally `log.Fatal` prints a message and then does an `os.Exit(1)` type thing so this will take a bit more refactoring

kfdm:
Yeah, after doing the first pass on just `main.go` I realized there will be quite a few other files that will need to be fixed up to use the same logging, but I will make sure to fix the spelling and error messages to handle OCD tendencies :)

beorn7:
Yeah, I guess the best is to pass the logger around. `func Delete(ms storage.MetricStore)` has to become `func Delete(ms storage.MetricStore, l log.Logger)`. Happy to hear other ideas.

main.go
beorn7: So you said you have a problem with this one. What's the nature of the problem?

kfdm: logger is defined within `func main()` (not global) so `func interruptHandler` can't see it. I either need to move logger as a global singleton (probably not wise) or refactor interuptHandler to get a reference to logger

beorn7: I see. Yeah, I guess giving the `func interruptHandler` an additional argument for the logger is the way to go.

kfdm:
Ended up touching more files than I initially expected, but hopefully I covered most things. I imagine some of the help strings will need a bit of an audit.

handler/status.go
```diff
@@ -22,11 +22,13 @@ import (
...
 	"time"
 
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"github.com/prometheus/common/log"
+
```
beorn7: So yeah, the convention hasn't even been followed precisely here. But you would make me very happy if you fixed it while adding the now imports. It should be: One block for standard imports, one for unnamed imports from other repos, one for named imports from other repos, one for unnamed imports from the same repo, and finally one for named imports from the same repo.

Thus, the now go-kit imports should be in this block. the kingpin import should be separate below, and again separate and at the bottom should be the pushgateway/storage import.

main.go
```diff
@@ -26,10 +26,13 @@ import (
...
 	"time"
 
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/julienschmidt/httprouter"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"github.com/prometheus/common/log"
+	"github.com/prometheus/common/promlog"
+	promlogflag "github.com/prometheus/common/promlog/flag"
```
beorn7: Put this named import in the same block as the `dto` import below. (Explanation see above comment.)

handler/status.go
beorn7: In different news: I'm not even sure we need to name the `kingpin` import (may be done as in `main.go`).

main.go
```diff
@@ -54,18 +57,20 @@ func main() {
...
 	kingpin.MustParse(app.Parse(os.Args[1:]))
 
+	logger := promlog.New(&promlogConfig)
 	*routePrefix = computeRoutePrefix(*routePrefix, *externalURL)
 
-	log.Infoln("Starting pushgateway", version.Info())
-	log.Infoln("Build context", version.BuildContext())
-	log.Debugf("Prefix path is '%s'", *routePrefix)
-	log.Debugf("External URL is '%s'", *externalURL)
+	level.Info(logger).Log("msg", "Starting pushgateway", "version", version.Info())
```
beorn7: Nit: lower case “starting”.

main.go
```diff
@@ -54,18 +57,20 @@ func main() {
...
+	level.Info(logger).Log("build_context", version.BuildContext())
+	level.Debug(logger).Log("msg", "Prefix path", "path", *routePrefix)
```
beorn7: Nit: lower case “prefix”.

main.go
```diff
@@ -54,18 +57,20 @@ func main() {
...
+	level.Debug(logger).Log("msg", "External URL", "url", *externalURL)
```
beorn7: Nit: lower case “external”.

main.go
```diff
@@ -89,36 +94,37 @@ func main() {
...
 	err = (&http.Server{Addr: *listenAddress, Handler: r}).Serve(l)
-	log.Errorln("HTTP server stopped:", err)
+	level.Error(logger).Log("msg", "HTTP server stopped", "err", err)
 	// To give running connections a chance to submit their payload, we wait
 	// for 1sec, but we don't want to wait long (e.g. until all connections
 	// are done) to not delay the shutdown.
 	time.Sleep(time.Second)
 	if err := ms.Shutdown(); err != nil {
-		log.Errorln("Problem shutting down metric storage:", err)
+		level.Error(logger).Log("msg", "Problem shutting down metric storage", "err", err)
```
beorn7: Nit: lower case “problem”.

main.go
```diff
@@ -156,10 +162,10 @@ func computeRoutePrefix(prefix string, externalURL *url.URL) string {
 	return prefix
 }
 
-func interruptHandler(l net.Listener) {
+func interruptHandler(l net.Listener, logger log.Logger) {
 	notifier := make(chan os.Signal, 1)
 	signal.Notify(notifier, os.Interrupt, syscall.SIGTERM)
 	<-notifier
-	log.Info("Received SIGINT/SIGTERM; exiting gracefully...")
+	level.Info(logger).Log("msg", "Received SIGINT/SIGTERM; exiting gracefully...")
```
beorn7: Nit: lower case “received”.

storage/diskmetricstore.go
```diff
@@ -22,9 +22,11 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+
```
beorn7: As above, no blank line and sort lexicographically in the common block.

storage/diskmetricstore.go
```diff
@@ -123,18 +128,15 @@ func (dms *DiskMetricStore) GetMetricFamilies() []*dto.MetricFamily {
...
+					level.Info(dms.logger).Log("msg", "metric famlies have inconsistent help strings", "new", mf, "old", existingMF)
 				}
 				// Type inconsistency cannot be fixed here. We will detect it during
 				// gathering anyway, so no reason to log anything here.
 				existingMF.Metric = append(existingMF.Metric, mf.Metric...)
 			} else {
 				copied := false
 				if help, ok := dms.predefinedHelp[name]; ok && mf.GetHelp() != help {
-					log.Infof("Metric family '%s' has the same name as a metric family used by the Pushgateway itself but it has a different help string. Changing it to the standard help string %q. This is bad. Fix your pushed metrics!", mf, help)
+					level.Info(dms.logger).Log("msg", "metric family has the same name as a metric family used by the Pushgateway itself but it has a different help string. Changing it to the standard help string. This is bad. Fix your pushed metrics!", "metric", mf, "help", help)
```
beorn7: Hmm, yeah, that's a special case where we have whole sentences. I would probably start with capital “Metric” in this case.

The logging key names should probably be the following to avoid misunderstandings:
- `"metric"` → `"metric_family"`
- `"help"` → `"standard_help"`

storage/diskmetricstore.go
```diff
@@ -190,12 +192,9 @@ func (dms *DiskMetricStore) loop(persistenceInterval time.Duration) {
...
 					persistStarted := time.Now()
 					if err := dms.persist(); err != nil {
-						log.Errorln("Error persisting metrics:", err)
+						level.Error(dms.logger).Log("msg", "error persisting metrics", "err", err)
 					} else {
-						log.Infof(
-							"Metrics persisted to '%s'.",
-							dms.persistenceFile,
-						)
+						level.Info(dms.logger).Log("msg", "Metrics persisted", "file", dms.persistenceFile)
```
beorn7: Nit: lower case “metrics”.

storage/diskmetricstore_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-kit/kit/log"
+
```
beorn7: as above, merge with following block.

kfdm:
Thank you for all the detailed feedback 🙇 
I think I have taken care of each comment.

beorn7:
Thanks for all your work and your understanding for my OCD. :o)

PR #265, opened 2019-07-02
Build ARM container images

simonpasquier:
It also updates the Circle CI configuration to use the Prometheus orb.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,9 @@
 ---
 version: 2.1
 
+orbs:
+  prometheus: prometheus/prometheus@0.1.0
+
 executors:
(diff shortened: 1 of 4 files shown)


simonpasquier:
It did work ;)

sabrehagen:
How can I access the pre-build containers for arm7? Currently I get the following on raspbian:

```sh
docker pull prom/pushgateway
Using default tag: latest
latest: Pulling from prom/pushgateway
no matching manifest for linux/arm/v7 in the manifest list entries
```

simonpasquier:
@sabrehagen it _should_ work without pulling the arch-specific image. Can you try using Quay (eg `docker pull quay.io/prometheus/pushgateway`)?

Issue #266, opened 2019-07-12
"Different help string " will be return  500 Internal Server Error 


imluyuan:
I use Wmi exporter from my win2008R and collect data use pushgateway.

I found error log from pushgateway.
```
time="2019-07-12T08:58:36Z" level=info msg="Metric family 'name:\"go_memstats_sys_bytes\" help:\"Number of bytes obtained by system. Sum of all system allocations.\" type:GAUGE metric:<label:<name:\"instance\" value:\"v-eb86f815-d6e6-4ebe-9fcd-6ec7b10ba95f-192.168.0.15\" > label:<name:\"job\" value:\"virt-by-push-for-windows\" > gauge:<value:1.16718e+07 > > ' has the same name as a metric family used by the Pushgateway itself but it has a different help string. Changing it to the standard help string \"Number of bytes obtained from system.\". This is bad. Fix your pushed metrics!" source="diskmetricstore.go:137"
```

When i push this metrics that service return 500 code
![image](https://user-images.githubusercontent.com/3294496/61118465-5d2c2000-a4cb-11e9-8c7f-7721e2d59240.png)

Delete this metrics that service return 200 
![image](https://user-images.githubusercontent.com/3294496/61118595-a3817f00-a4cb-11e9-85fa-79a58d9e9201.png)

beorn7:
The info log message says that the help string was adjusted. Inconsistent help strings don't cause this kind of problem anymore.

I assume there is something else broken with that group of pushed metrics.

If you go to the /metrics endpoint of the Pushgateway with your browser, you should see an error message (which we probably should also log).

beorn7:
https://github.com/prometheus/pushgateway/pull/267 will switch on logging for this kind of error.

imluyuan:
Nice ~ 
This problem has been resolved.
@beorn7  welcome to Beijing，Winner winner, chicken dinner！

beorn7:
With #267 merged, I'll close this. (Will also do a release this week.)

PR #267, opened 2019-07-12
Also log error messages during gathering

beorn7:
Quite a few of users have run into the problem that they see a scrape
error in the Prometheus UI but don't see the error. (It's not
displayed in the Prometheus UI, you have to go to the /metrics
endopint yourself to see it.) Logging it makes the problem much more obvious, because all those users looked at what the PGW is logging, which was nothing in that case.

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"fmt"
 	"net"
 	"net/http"
 	"net/http/pprof"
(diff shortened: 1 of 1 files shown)


beorn7:
#266 is an example for a confused user.

beorn7:
Confusion is even higher here because the minor issue of the different HELP string _is_ logged while the actual fatal error is only sent as the body of the 500 response.

beorn7:
Wait, this works differently now with gokit logging. Fix incoming...

beorn7:
@grobie gentle ping…

beorn7:
Thanks @gouthamve (@grobie I got hold of @gouthamve in person, so I convinced him to do this fairly simple PR).

PR #268, opened 2019-07-19
 Add the option to base64-encode label names in push URLs

### LINKED ISSUE #97: If parameters include '/' ,It will response code 400
If tags parameters include '/'  ,It will response code 400.The '/' is deal with URLEncoder. How to do?

beorn7:
This fixes #97.

I marked the maintainers of the official client libraries as reviewers, as you are most directly affected by this.

This still needs test added, but I want to get your thoughts first before writing them.

Copied from the commit description:

I did quite a lot of research about how to encode `/` in RESTful
request URLs. There doesn't seem to be a general consensus how to do
it (besides just not to do it). I found a few cases where systems
would accept base64 encoding (which has to be the "URL and Filename
safe" variant, otherwise you have `/` in there again, but at least the
existence of this variant in RFC 4648 tells us that it is kind of meant
for what we are doing here). But then the next problem arises: How to
mark what is base64 encoded and what is plain text? In most cases I
saw, the respective field is just marked as being always base64
encoded, which would be sad for our case as there are very few
situations where you actually need base64 encoding. One case used a
marker like `$base64:...`, which I found quite weird. However, I think
using some kind of marker is a good idea. I picked a leading `=` here
as that's a character that is rarely used as a first character of a
label value (and hopefully even less so for the label value in a
grouping key). It is also used in PGP to prefix a base64 encoded string.

I put details in the documentation (see README.md in this commit).

As illustrated there, a nice byproduct is that the encoding of special
characters is a bit saner now (imagine Chinese letters in label
values, which might be pretty common in, well, China).

As base64 encoding is readily available in standard libraries and even
as command line tools (again see examples in the documentation), I
think this is a pretty doable and not overly invasive way of dealing
with the problem.

Client libraries should of course support this encoding transparently.

The competition is mostly going away from the RESTful style and use
URL parameters. If we completely switch to it, we would break
everything hard. If we offered it as an alternative, we would deviate
from our best practice of only offering one way of doing one thing. It
would be a quite invasive change either way.

We could also use a custom escaping scheme just for the case of `/`,
but there is no precedent for that. We would have a solution only used
for the Pushgateway, and everyone had to implement that escaping
scheme on their own. (base64 is a more complex encoding, but the
tooling is readily available everywhere.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,24 +211,24 @@ All pushes are done via HTTP. The interface is vaguely REST-like.
 
 The default port the push gateway is listening to is 9091. The path looks like
 
-    /metrics/job/<JOBNAME>{/<LABEL_NAME>/<LABEL_VALUE>}
+    /metrics/job/<JOB_NAME>{/<LABEL_NAME>/<LABEL_VALUE>}
 
-`<JOBNAME>` is used as the value of the `job` label, followed by any
+`<JOB_NAME>` is used as the value of the `job` label, followed by any
 number of other label pairs (which might or might not include an
 `instance` label). The label set defined by the URL path is used as a
 grouping key. Any of those labels already set in the body of the
 request (as regular labels, e.g. `name{job="foo"} 42`)
 _will be overwritten to match the labels defined by the URL path!_
 
-If the job name or any label value starts with a `=`, the remaining part is
-interpreted as a base64 encoded string according to [RFC 4648, using the URL
-and filename safe
+If `job` or any label name is suffixed with `@base64`, the following job name
+or label value is interpreted as a base64 encoded string according to [RFC
+4648, using the URL and filename safe
 alphabet](https://tools.ietf.org/html/rfc4648#section-5). (Padding is
 optional.) This is the only way of using job names or label values that contain
-a `/` or start with a `=`. For other special characters, the usual URI
-component encoding works, too, but the base64 might be more convenient.
+a `/`. For other special characters, the usual URI component encoding works,
+too, but the base64 might be more convenient.
 
-Ideally, client libraries take care of this encoding.
+Ideally, client libraries take care of the suffixing and encoding.
 
 Examples:
 
@@ -237,18 +237,17 @@ Examples:
 
 	  /metrics/job/directory_cleaner/path//var/tmp
 	  
-  Instead, use the base64 URL-safe encoding with a leading `=`:
+  Instead, use the base64 URL-safe encoding for the label value and mark it by
+  suffixing the label name with `@base64`:
   
-  	  /metrics/job/directory_cleaner/path/=L3Zhci90bXA
+  	  /metrics/job/directory_cleaner/path@base64/L3Zhci90bXA
 	  
   If you are not using a client library that handles the encoding for you, you
   can use encoding tools. For example, there is a command line tool `base64url`
   (Debian package `basez`), which you could combine with `curl` to push from
   the command line in the following way:
   
-      echo 'some_metric{foo="bar"} 3.14' | curl --data-binary @- http://pushgateway.example.org:9091/metrics/job/directory_cleaner/path/=$(echo -n '/var/tmp' | base64url)
- 
-  Note the leading `=`.
+      echo 'some_metric{foo="bar"} 3.14' | curl --data-binary @- http://pushgateway.example.org:9091/metrics/job/directory_cleaner/path@base64/$(echo -n '/var/tmp' | base64url)
   
 * The grouping key `job="titan",name="Προμηθεύς"` can be represented
   “traditionally” with URI encoding:
@@ -257,7 +256,7 @@ Examples:
 	  
   Or you can use the more compact base64 encoding:
   
-      /metrics/job/titan/name/=zqDPgc6_zrzOt864zrXPjc-C
+      /metrics/job/titan/name@base64/zqDPgc6_zrzOt864zrXPjc-C
 
 ### `PUT` method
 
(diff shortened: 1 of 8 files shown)


beorn7:
@Sinjo and @dmagliola , feel nominated as reviewers, too. I just cannot put you in the formal list above as you are not collaborators in this repo.

beorn7:
I had a minor-correction commit mixed into this. It's now confusing if you want to see what changed in the README.md. I'll push that to master directly so that we get rid of it here.

brian-brazil:
This may break existing users who have an = in their label value, how about a new endpoint?

beorn7:
> This may break existing users who have an = in their label value,

That's why I want to introduce it before releasing 1.0.

> how about a new endpoint?

I would grudgingly add a new endpoint if we were post 1.0. I don't want to introduce that additional complication for that unlikely case in a pre-1.0 software.

beorn7:
I have an idea how to avoid both the breaking change _and_ adding a separate endpoint: Put the marker character in the label name, not the label value. In that way, we can use something that would otherwise be illegal in a label name. For example, appending `$base64` would read quite nicely. The examples from the README.md would look like the following: 

`/metrics/job/directory_cleaner/path//var/tmp` (doesn't work syntactically) → `/metrics/job/directory_cleaner/path$base64/L3Zhci90bXA`

`/metrics/job/titan/name/Προμηθεύς` (needs to be URL-encoded) → `/metrics/job/titan/name$base64/zqDPgc6_zrzOt864zrXPjc-C`

I like that a lot.

brian-brazil:
Yeah, that's better.

In OM terms, we'd talked about having a special target_info metric that'd contain target labels for push cases (to avoid users putting them on every metric, and messing things up for top-down systems) but it's not been fleshed out.

handler/delete.go
```diff
@@ -37,7 +38,12 @@ func Delete(ms storage.MetricStore, logger log.Logger) func(http.ResponseWriter,
 	instrumentedHandler := promhttp.InstrumentHandlerCounter(
 		httpCnt.MustCurryWith(prometheus.Labels{"handler": "delete"}),
 		http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
-			job := ps.ByName("job")
+			job, err := maybeDecodeBase64(ps.ByName("job"))
+			if err != nil {
+				http.Error(w, fmt.Sprintf("invalid base64 encoding in job name %q: %v", ps.ByName("job"), err), http.StatusBadRequest)
+				level.Debug(logger).Log("msg", "invalid base64 encoding in job name", "job", ps.ByName("job"), "err", err.Error())
```
brian-brazil: Should this be higher than debug?

beorn7: The idea here is this is an error of the pushing client (not an error in the PGW), and the error message is already sent back in the HTTP response to the pushing client. We are not really doing query logging in the PGW, and this would be more query logging.

However, I do not feel strongly about this distinction. However, if we decided to log this at a higher level, we should change it everywhere else where similar things are done (e.g. two more instances here in the very same handler function and many more in the push handler). We should also log more then (like IP number of the pushing client and such). My concern then would be that we are dealing with a different issue then the one this PR is about, and we should do that in a separate PR.

beorn7:
I'm implementing my proposed alternative solution now.

beorn7:
> In OM terms, we'd talked about having a special target_info metric that'd contain target labels for push cases (to avoid users putting them on every metric, and messing things up for top-down systems) but it's not been fleshed out.

Doesn't the grouping key fulfill that function already?

handler/delete.go
beorn7: As a data point: When we introduced those log lines, the decision to do it at debug level was deliberate. We can, of course, revisit that decision. Just saying that it is not an accident.

brian-brazil:
> Doesn't the grouping key fulfill that function already?

The idea was more that this would replace the grouping key, and more generally make target labels work for push in a way that doesn't break pull.

beorn7:
I don't quite get the difference between that and the existing grouping key, but we don't need to have that discussion now. Just wanted to make sure you thought about it.

beorn7:
This is now ready for review.

brian-brazil:
Yes, we had thought about it. I brought it up as it's related (and was a potential solution to this issue in passing).

README.md
```diff
@@ -211,19 +211,54 @@ All pushes are done via HTTP. The interface is vaguely REST-like.
...
+  	  /metrics/job/directory_cleaner/path$base64/L3Zhci90bXA
+	  
+  If you are not using a client library that handles the encoding for you, you
+  can use encoding tools. For example, there is a command line tool `base64url`
+  (Debian package `basez`), which you could combine with `curl` to push from
+  the command line in the following way:
+  
+      echo 'some_metric{foo="bar"} 3.14' | curl --data-binary @- http://pushgateway.example.org:9091/metrics/job/directory_cleaner/path\$base64/$(echo -n '/var/tmp' | base64url)
+    
+  Remember to escape the `$` in `$base64`.
```
brian-brazil: Hmm, I wonder if we should use a different character to avoid this problem.

beorn7: I mostly picket it because JetBrains apparently use `$base64:` somewhere in their restful APIs as a marker. But that's far from being a de-facto standard. Our marker character should just be one that doesn't need fancy URL-encoding. `@` was my second candidate on the list. Would you like that more?

brian-brazil: Yeah, @ would cause fewer problems.

beorn7: OK, changing it and will merge after.

beorn7:
@SuperQ @Sinjo @dmagliola as this is now strictly non-breaking, I take the liberty to not wait for your feedback before merging. However, if you have any concerns, please still let me know, so that they can be addressed before we have a PGW v1.

PR #271, opened 2019-07-23
Correctly remove the whole panel from the UI after deletion

### LINKED ISSUE #270: UI: Expanded group not deleted from UI
v0.8.0: If a visually expanded metric group is deleted via UI, the expanded part is not deleted from the view. (Reload fixes it. This is just a presentation issue.)

beorn7:
The structure changed a lot due to the Bootstrap upgrade. Now we have
to delete one level up.

Fixes #270.


brian-brazil:
:+1:

Issue #273, opened 2019-07-23
`--web.route-prefix` and `--web.web.external-url` still don't work exactly like in Prometheus


siebenmann:
## Bug Report
**What did you do?**

Looked at the web UI for Pushgateway 0.9.0 (in our case, this web page is behind an Apache reverse proxy).

**What did you expect to see?**

As in the 0.8.0 web UI, I expected to see icons for expanding or compacting the various pushed groups.

**What did you see instead? Under which circumstances?**

Ordinary characters instead of icons. I've attached a screenshot.

**Environment**

This has reproduced in Firefox 68.0.1 on Fedora Linux 29 and Windows 10, and Chrome 75.0.3770 on Fedora Linux 29. All environments are using uBlock Origin, but both uBlock Origin's request logger and Firefox's Network console show no blocked requests. Firefox also shows that the glyphicons-halflings-regular.woff2 font was fetched, and there are no errors in the console.

* System information:

   Pushgateway server (Ubuntu 18.04): Linux 4.15.0-52-generic x86_64

* Pushgateway version:

   ```
   pushgateway, version 0.9.0 (branch: HEAD, revision: 44d7ae6d9fb05dfe141eefb4bdf1bed7b89dee31)
     build user:       root@45c30774a08c
     build date:       20190723-15:31:55
     go version:       go1.12.7
   ```

* Pushgateway command line:

   ``/opt/prometheus/pushgateway-0.9.0.linux-amd64/pushgateway --web.listen-address=127.0.0.1:9091 --persistence.file=/var/local/prometheus/pushgateway/persist.data``

* Logs: Pushgateway is not reporting any log messages that seem relevant. The only log output since restarting with 0.9.0 is messages about persisting metrics.

Screenshot of what this looks like:
![Screenshot_2019-07-23 Prometheus Pushgateway](https://user-images.githubusercontent.com/865382/61735813-b6f5db00-ad52-11e9-80fc-7919dc91aa08.png)

beorn7:
I have seen the some once, and a deep reload (shift-F5, ctrl-F5, or whatever your browser/OS uses for a deep reload) fixed it. There are cache busters in the resource URLs now, but it still sometimes happens somehow.

siebenmann:
This happens for me even after a deep reload, and even in a browser instance that has no caches (it's a setup I use for testing, where I always want to start from a clean slate). As a double check, I started Firefox with a completely new scratch `$HOME` and still saw the problem with our Pushgateway setup.

[...]

PR #274, opened 2019-07-25
Next attempt to make the --web.* flags work like in Prometheus

beorn7:
This also straightens out a few other things.

Note: I'm once more fairly convinced that we don't really need the `web.external-url` flag at all. We should be fine with just `web.route-prefix`. (But then the latter needed to be set in exactly the other case as for the Prometheus server. So perhaps we leave it as is to not confuse users. At some point, we might need `web.external-url` after all, for the same reasons Prometheus needs it.)

### DIFF
--- a/handler/status.go
+++ b/handler/status.go
@@ -27,7 +27,6 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"gopkg.in/alecthomas/kingpin.v2"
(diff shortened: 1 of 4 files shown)


beorn7:
This should fix #273 .

beorn7:
@juliusv could you have a look? You are most familiar with the web stuff in prometheus/prometheus, and this is supposed to emulate what Prometheus is doing. @siebenmann has tried this change in a production setup that didn't work before, and with this change, all seems to work exactly like with Prometheus.

beorn7:
Thanks.

Issue #264, opened 2019-06-26
Add an endpoint and UI button to wipe the storage completely


beorn7:
If too much garbage has been pushed, you might want to erase the whole storage and start from a clean slate. Currently, you would do the following:
- shut down the PGW
- delete the persistence file
- start PGW again

It would be easy to add an endpoint like `/-/reset` or `/-/wipe`, only active if a (yet to be added) flag ~~`--web.enable-lifecycle`~~ `--web.enable-admin-api` is set (similar to #136).

If that behavior is activated, it would also be good to add a button to the UI “Delete all groups”.

brian-brazil:
This would be an admin endpoint rather than lifecycle. We should also have it as PUT rather than POST, for XSRF reasons.

beorn7:
Ah right, the Prometheus server has `--web.enable-admin-api`, should be the same here.

snebel29:
I would like to work on this feature, I will take a look to the code and comment back

snebel29:
This is the current TODO list for this feature request

- [x] Add API call to wipe the storage completely
- [ ] Document `/wipe` API call
- [ ] Add button to delete all metric groups

beorn7:
This is done. I'll cut a release at my next convenience so that the general public can make use of it.
\o/

Issue #184, opened 2018-05-30
Expose metrics metadata in HTTP API


pbadenski:
The way I understand it when interacting with a registered metric via pushgateway API (PUT, POST, DELET, etc.) you need to match grouping key exactly. I also believe that the only way to figure out the grouping key for metric is from pushgateway UI (https://github.com/prometheus/pushgateway/blob/master/handler/status.go#L85 seems to be doing it). This becomes especially relevant when trying to remove metrics from pushgateway. 

Building on top of that - it would be very helpful to have this information available through HTTP API in a machine readable format (JSON?). Ideally combined with other information being reported in the UI.

snebel29:
The above link to `status.go#L85` seems to be not relevant anymore since is pointing to master which changed over time, I think it's fair to assume that the below line was probably the referred one

https://github.com/prometheus/pushgateway/blob/231071b3e0a57b58df717bba5f4e5d88cac95389/handler/status.go#L85

My understanding is that what we want to do here is

1. Provide a programmatic way of retrieving metric information, including its grouping key. Something like allow `GET /metrics` and returning results back, may be in pages of N metrics => To verify if Prometheus is doing something similar, to match its behavior.
2. Adapt the UI to use the new GET endpoint instead of relying on templating them.

Those could come in separate PRs, am I right?

If there is no opposition I'm happy to take this issue.

Cheers

beorn7:
Yes, that's broadly correct. `/metrics` is probably not a good endpoint as it is used for the exposition of (all the) metrics in the Prometheus exchange format. It should be an endpoint like `/api/v1/...` (where the `...` can certainly be `metrics` – best to see what the Prometheus server is doing there). Pagination would be great, but it doesn't have to be in the initial solution (we don't have pagination in the status page right now, and pushing an excessive amount of metrics to the PGW has a smell anyway).

snebel29:
Great, will work on this next week.

snebel29:
Hi @beorn7,

I took a brief look to the [Prometheus querying API](https://prometheus.io/docs/prometheus/latest/querying/api/) and found:

- API Endpoints are all exposed on `/api/vN/` path.
- Returns JSON using the [following](https://prometheus.io/docs/prometheus/latest/querying/api/#format-overview) format/structure.
- There is multiple endpoints to get metrics data and time series, using different filters and matching patterns (query, range, series, etc.).
[...]

Issue #149, opened 2018-03-23
Move to common/route package


beorn7:
https://github.com/prometheus/common/tree/master/route is most likely applicable to the Pushgateway, too. Note the discussion in #146 about caveats. Also, see https://github.com/prometheus/prometheus/issues/4004 .

beorn7:
Now also note https://github.com/prometheus/prometheus/pull/4256 . I'm confused now what the flag means for the Prometheus server. This needs clarification.

beorn7:
To learn about the meaning of `--web.external-url` and `--web.route-prefix`, see @brian-brazil 's helpful blog posts
- https://www.robustperception.io/using-external-urls-and-proxies-with-prometheus/
- https://www.robustperception.io/external-urls-and-path-prefixes/

My understanding so far is that we don`t need `--web.external-url` as we only ever work with relative paths. But for one, I might be wrong, and even if I'm right, this might change eventually.

brian-brazil:
Route prefix is still needed to handle where there's a prefix on the incoming http requests that's not the same as on the external-url. External URL is required as the HTML uses absolute paths.

beorn7:
I think that's what I said, no?

brian-brazil:
I interpreted it as you saying external-url was not needed, but you were unsure. I said that we do need external-url with the current codebase.

beorn7:
So you are saying my assumption that PGW only uses relative paths is wrong?

brian-brazil:
Yes, from the current html:

```
    <script src="/static/functions.js"></script>
```

beorn7:
Hmm, but we are not using the host name anywhere, do we? Shouldn't we then be fine with the route-prefix?

brian-brazil:
The slash at the start makes it absolute, without that slash it'd be okay. Otherwise the JS won't load if the browser is using a different prefix to what the pushgateway is configured for.

beorn7:
I think I got it now.

I have still conflated the route-prefix implementation currently in the PGW with the one in the Prometheus server. The former is essentially external-url, but without the host and schema part (as we don't need that one), while the latter doesn't exist in PGW (but would be required to handle the case where the reverse proxy rewrites paths).

With all that confusion, I'm thinking what is actually keeping us from making all paths relative, at least in the PGW, and get rid of the whole prefix and external url stuff altogether?

brian-brazil:
[...]

PR #283, opened 2019-09-05
Feature: wipe storage completely

snebel29:
This PR relates to https://github.com/prometheus/pushgateway/issues/264

I've created a POC so this PR would still be a working progress, I just wanted to get early feedback before keep moving forward, summary:

- Added the new `--web.enable-admin-api` flag
- Added the new `/-/wipe` route for the `PUT` verb
- I've created a new `WipePersistentFile` handler, this handler have a bunch of comments and potential TODOs.

Once and assuming that at some point the mentioned above is ok, still would be necessary to

- [ ] Implement wipe button in the UI
- [ ] Add test for at least the new wipe handler
- [ ] Rebase/Squash

I hope it makes sense, but please let me know if it doesn't...

Thanks!

### DIFF
--- a/handler/handler_test.go
+++ b/handler/handler_test.go
@@ -47,6 +47,10 @@ func (m *MockMetricStore) GetMetricFamiliesMap() storage.GroupingKeyToMetricGrou
 	panic("not implemented")
 }
 
+func (m *MockMetricStore) Wipe() error {
+	return nil
+}
+
 func (m *MockMetricStore) Shutdown() error {
 	return nil
 }
--- a/handler/handler_test.go
+++ b/handler/handler_test.go
@@ -47,8 +47,7 @@ func (m *MockMetricStore) GetMetricFamiliesMap() storage.GroupingKeyToMetricGrou
 	panic("not implemented")
 }
 
-func (m *MockMetricStore) Wipe() error {
-	return nil
+func (m *MockMetricStore) Wipe() {
 }
 
 func (m *MockMetricStore) Shutdown() error {
--- a/handler/handler_test.go
+++ b/handler/handler_test.go
@@ -39,6 +39,10 @@ func (m *MockMetricStore) SubmitWriteRequest(req storage.WriteRequest) {
 	m.lastWriteRequest = req
 }
 
+func (m *MockMetricStore) GetPersistenceFile() string {
+	return ""
+}
+
 func (m *MockMetricStore) GetMetricFamilies() []*dto.MetricFamily {
 	panic("not implemented")
 }
--- a/handler/handler_test.go
+++ b/handler/handler_test.go
@@ -62,14 +62,8 @@ func (m *MockMetricStore) Ready() error {
 	return nil
 }
 
-func newMockMetricStore() MockMetricStore {
-	return MockMetricStore{
-		metricGroups: storage.GroupingKeyToMetricGroup{},
-	}
-}
-
 func TestHealthyReady(t *testing.T) {
-	mms := newMockMetricStore()
+	mms := MockMetricStore{}
 	req, err := http.NewRequest("GET", "http://example.org/", &bytes.Buffer{})
 	if err != nil {
 		t.Fatal(err)
@@ -91,7 +85,7 @@ func TestHealthyReady(t *testing.T) {
 }
 
 func TestPush(t *testing.T) {
-	mms := newMockMetricStore()
+	mms := MockMetricStore{}
 	handler := Push(&mms, false, false, logger)
 	handlerBase64 := Push(&mms, false, true, logger)
 	req, err := http.NewRequest("POST", "http://example.org/", &bytes.Buffer{})
@@ -395,7 +389,7 @@ another_metric{instance="baz"} 42
 }
 
 func TestDelete(t *testing.T) {
-	mms := newMockMetricStore()
+	mms := MockMetricStore{}
 	handler := Delete(&mms, false, logger)
 	handlerBase64 := Delete(&mms, true, logger)
 
@@ -566,10 +560,11 @@ func TestWipeMetricStore(t *testing.T) {
 	// so they can be returned by GetMetricFamiliesMap() to later send write
 	// requests for each of them.
 	metricCount := 5
-	mms := newMockMetricStore()
+	mgs := storage.GroupingKeyToMetricGroup{}
 	for i := 0; i < metricCount; i++ {
-		mms.metricGroups[uint64(i)] = storage.MetricGroup{}
+		mgs[uint64(i)] = storage.MetricGroup{}
 	}
+	mms := MockMetricStore{metricGroups: mgs}
 
 	// Wipe handler should return 202 and delete all metrics.
 	wipeHandler := WipeMetricStore(&mms, logger)
--- a/handler/handler_test.go
+++ b/handler/handler_test.go
@@ -44,7 +44,7 @@ func newMockMetricStore() MockMetricStore {
 }
 
 // This mock method just set lastWriteRequest and write metrics, but some fields
-// are set to its zero value
+// are set to its zero value.
 func (m *MockMetricStore) SubmitWriteRequest(req storage.WriteRequest) {
 	m.lastWriteRequest = req
 
@@ -56,7 +56,7 @@ func (m *MockMetricStore) SubmitWriteRequest(req storage.WriteRequest) {
 	}
 
 	// Update metric contains similar logic as you find
-	// within diskmetricstore.processWriteRequest()
+	// within diskmetricstore.processWriteRequest().
 	for name, _ := range req.MetricFamilies {
 		group, ok := m.metricGroups[key]
 		if !ok {
@@ -78,7 +78,7 @@ func (m *MockMetricStore) GetMetricFamilies() []*dto.MetricFamily {
 }
 
 func (m *MockMetricStore) GetMetricFamiliesMap() storage.GroupingKeyToMetricGroup {
-	// Simply return them without performing a copy
+	// Simply return them without performing a copy.
 	return m.metricGroups
 }
 
@@ -107,7 +107,7 @@ func TestWipeMetricStore(t *testing.T) {
 	}
 	w := httptest.NewRecorder()
 
-	// Push a few metrics to the MetricStore
+	// Push a few metrics to the MetricStore.
 	count := 10
 	for i := 0; i < count; i++ {
 		pushHandler(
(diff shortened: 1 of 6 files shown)


brian-brazil:
> Added the new --web.enable-admin-api flag

Is this needed given this is already possible with the existing API?

beorn7:
> Is this needed given this is already possible with the existing API?

The existing API needs to have the grouping key in the URL. It's much less likely to accidentally delete everything. Therefore, I like the idea to make the "wipe everything" endpoint opt-in.

And yes, it should follow the same pattern as prometheus/prometheus, i.e. `admin` in the URL, cf. https://prometheus.io/docs/operating/security/#prometheus

handler/wipe.go
```diff
@@ -0,0 +1,51 @@
...
+		// if there was an error persisting the last metrics, still I can see some value in logging
+		// potential errors for troubleshoting
+		if err := dms.Shutdown(); err != nil {
+			level.Error(logger).Log("msg", "problem shutting down metric storage", "err", err)
+		}
+
+		n := newStorage()
+		// Delete persitence file
+		// TODO: should we handle the error from os.Remove? What exactly should we be doing?
+		os.Remove(n.GetPersistenceFile())
```
beorn7: I think it is a bit low-level to ask the storage for a file to delete, delete it with `os.Remove` and then also rely on this code to create a new storage and not to keep using the old one with the deleted file.

How about giving the storage a `Wipe()` method that just does the stuff under the hood?

At some point, people might want to introduce storage implementation that use other backends than a file in a local file system.

handler/wipe.go
```diff
@@ -0,0 +1,51 @@
...
+	"net/http"
+	"os"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+
+	"github.com/prometheus/pushgateway/storage"
+)
+
+func WipePersistentFile(
```
beorn7: I'd consistently use `Persistence` as before. You are now using `Persistent` at several places.

handler/wipe.go
beorn7: In fact, I think you shouldn't be concerned with deleting the file itself. (If the user hasn't requested persistence, there won't be any, but wiping of the (in-memory) storage should still be possible.)

The proposed `Wipe()` method should just iterate through everything there is in storage, delete it, and then let the usual persistence mechanism take care of writing that state to disk.

snebel29: Yeah actually it makes sense

handler/wipe.go
snebel29: Good catch, will do

snebel29:
I will also add admin to the URL path to be consistent

snebel29:
@beorn7 I think now this pretty much what you suggested, I still want to 

1. Explore the `/admin` path in Prometheus code base to ensure the path `/api/v1/admin` and just returning status code 202 (no content) is consistent with the existing admin endpoints in the ecosystem.

```
$ curl -XPUT http://127.0.0.1:9091/api/v1/admin/wipe -vvv
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 9091 (#0)
> PUT /api/v1/admin/wipe HTTP/1.1
> Host: 127.0.0.1:9091
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 202 Accepted
< Date: Sat, 07 Sep 2019 10:44:57 GMT
< Content-Length: 0
< 
* Connection #0 to host 127.0.0.1 left intact
```

2. Add a test case for the `DiskMetricStore.Wipe()` method
3. Add `Wipe All Metric Groups` button in the UI, I think above the existing `Delete Group` button under metrics.

What do you think?

Thanks in advance

main.go
```diff
@@ -110,6 +111,12 @@ func main() {
 		}),
 	)
 
+	if *enableAdminAPI {
+		// TODO: I'm not fully clear whether this is the desired admin path,
+		//  from prometheus codebase I would say so, but I will double check.
+		r.Handler("PUT", *routePrefix+"/api/v1/admin/wipe", handler.WipeMetricStore(ms, logger))
```
beorn7: The path looks kind of clunky, but mostly because we don't have any other API calls yet. However, I'm sure we'll got more of those, see #184 , so it's probably best to use this URL to be both consistent with Prometheus and able to version the API if needed.

storage/diskmetricstore.go
```diff
@@ -100,6 +100,17 @@ func (dms *DiskMetricStore) SubmitWriteRequest(req WriteRequest) {
...
 }
 
+// Wipe implements the MetricStore interface.
+func (dms *DiskMetricStore) Wipe() {
+	// Delete all metric groups by sending write requests with MetricFamilies equal to nil
+	for _, group := range dms.GetMetricFamiliesMap() {
+		dms.SubmitWriteRequest(WriteRequest{
+			Labels:    group.Labels,
+			Timestamp: time.Now(),
+		})
```
beorn7: This implementation is so nice and elegant and it only uses other methods from the interface (rather than any implementation details) of the MetricStore. Thus, I have second thoughts now if we really want a `Wipe` method in the interface. This code could as well live in `wipe.go`. What do you think?

main.go
snebel29: Yeah, I had the same feeling. OK we keep it then

storage/diskmetricstore.go
snebel29: It's a good point, my two cents follow

Since it's possible to wipe the entire MetricStore (no matter whether this is the Disk implementation or any other future MetricStore implementation) by simply using already existing methods from the interface, we could avoid forcing future MetricStore implementations to implement their own Wipe() method, which would most likely be just the same code.

Keeping Wipe() method within the interface would make sense having plans to call it from within the MetricStore, it'd be nice if we could somehow and "elegantly" mixin a generic Wipe implementation method without forcing to each implementation to create its own and having it available through a pointer receiver method, but I'm not sure if this is possible.

beorn7: The latter could still be addressed by refactoring the code into a helper function, should the need ever come up.

snebel29: OK, let's move the wipe logic to the handler then

handler/wipe.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	"github.com/prometheus/pushgateway/storage"
+)
+
+// WipeMetricStore deletes all the metrics in MetricStore
```
beorn7: Nit: Sentence should end with a period. (Applies to other comments, too. We have a tradition of proper punctuation in the project. ;o)

snebel29: :+1:

snebel29:
I followed your suggestions, how does it looks like now?

Just one personal observation

There is a substantial amount of relevant code not being tested within `main.go`, feels like this all might have started simple and skipped the tests, but I think at the current state this would be great to have a good refactor for allowing testing and may be using same components as other projects, example `prometheus/common/route`

I start looking for adding the button to the UI, I will share the approach before start coding and request for feedback

handler/handler_test.go
```diff
@@ -59,8 +62,14 @@ func (m *MockMetricStore) Ready() error {
 	return nil
 }
 
+func newMockMetricStore() MockMetricStore {
```
beorn7: Not a big deal, but in this scenario where the usage of the MockMetricStore is purely internal, and all what this constructor does is to create an empty map, I'd prefer to keep things simple and not have a constructor function.

handler/handler_test.go
```diff
@@ -551,3 +560,40 @@ func TestSplitLabels(t *testing.T) {
...
+
+func TestWipeMetricStore(t *testing.T) {
+	// Create MockMetricStore with a few GroupingKeyToMetricGroup metrics
+	// so they can be returned by GetMetricFamiliesMap() to later send write
+	// requests for each of them.
+	metricCount := 5
+	mms := newMockMetricStore()
+	for i := 0; i < metricCount; i++ {
+		mms.metricGroups[uint64(i)] = storage.MetricGroup{}
+	}
```
beorn7: Adding to my comment above, I'd write this  in the following way to avoid the need for a constructor:

```Go
	metricCount := 5
	mgs := storage.GroupingKeyToMetricGroup{}
	for i := 0; i < metricCount; i++ {
		mgs[uint64(i)] = storage.MetricGroup{}
	}
	mms := MockMetricStore{metricGroups: mgs}
```

In total, that's fewer lines (no constructor) and makes things more explicit.

beorn7:
> I start looking for adding the button to the UI, I will share the approach before start coding and request for feedback

It might make sense to do this in a later PR. Better to get this in and proceed in smaller steps. (For example, I have some work in a branch that will create merge conflicts with this PR. We should reconcile both better sooner than later.)

I'd say once my last nitty comment is addressed, we merge this.

beorn7:
About the test coverage in `main.go`: I'd see `main.go` as mostly boilerplate code, declarative in character, without a lot of complex logic. Surely, a lot could go wrong here, like typos in the routing components and such, but those bugs will get noticed very soon, and test code would boil down to "please enter this routing component again" (like those web forms that ask twice for your email address – similar to them, you could consistently mistype in both cases so that the test doesn't catch the bug anyway).

I'm not against improving test coverage, but I'd probably try to move complex logic out of `main.go` into a package, where it can be tested there, as a real unit test.

The worst offender in this case is `computeRoutePrefix`, more or less copied out of common/route. See #149 for the discussion if we should use that package here. After having fixed the `--web.external-url` and `--web.route-prefix` flags, I tend to say we should not move to it because it is mostly overkill for the relatively simple use case of the PGW.

handler/handler_test.go
snebel29: :+1:

handler/handler_test.go
snebel29: :+1:

snebel29:
After the last push, this is ready to merge :crossed_fingers: 

Thanks for your patience and for providing extended comments with the rationale behind your requests, that definitively helped.

beorn7:
Many thanks. And looking forward to your next contribution (e.g. adding the wipe button).

beorn7:
Another thing that still needs to be added is some documentation in the README.md.

snebel29:
Oh! very true, will do in another PR tomorrow

brian-brazil:
The security doc will also need a small update.

snebel29:
Thanks I will review it and update accordingly, tracking it on https://github.com/prometheus/pushgateway/pull/285

abelorosz:
Can I ask when are you planning to make a new release which includes this change?

beorn7:
We need #287 to complete, and then I can cut a new release.

PR #284, opened 2019-09-09
Update dependencies

beorn7:
Also, add a go version line, as this tends to be added automatically
by tools, but then it adds the Go version currently used, while
Pushgateway should just build fine with older Go versions. (Specifying
anything earlier than Go1.11 doesn't make sense, though, as Go modules
weren't supported previously.)


beorn7:
We still need to update golang-builder to Go1.13.

simonpasquier:
golangci-lint is broken with Go 1.13 (see https://github.com/golangci/golangci-lint/issues/652 and https://github.com/golangci/golangci-lint/issues/659). We also need https://github.com/prometheus/golang-builder/pull/76 which I haven't merged yet because of that same issue.

beorn7:
That explains it. Thanks. So we'll have more things waiting for getting unblocked.

beorn7:
Also needed to update the vendored files.

beorn7:
@simonpasquier build is green now.

PR #287, opened 2019-09-12
Add wipe UI button

snebel29:
This relates to https://github.com/prometheus/pushgateway/issues/264

For the moment this is just work in progress exploratory code, in order to get early feedback while fighting with bootstrap classes to get something that looks good.

So far the button works, I'm not completely happy with the grey `div.card` background of the button and I'm trying to remove it, also I have the feeling the colors of the new button are slightly different than the existing ones.

pushgateway-wipe-button

> Disclaimer: I really don't have much experience in front-end development, although I'm enjoying it.

Thanks


beorn7:
Disclaimer: Not a front-end developer, either.

I like it in general. You should, in any case, do not display the button at all if `--web.enable-admin-api` is not set.

snebel29:
Ah, good point thanks, I will be off for a few days but will try to finish
this up once back 🤙

snebel29:
@beorn7 I have the feeling the button is ready, squashing and accepting the new timestamps changes in conflicting file `asset/assets_vfsdata.go` should be enough to merge, what do you think?

Thanks

beorn7:
Cool. Will have a look ASAP. (Currently on vacation, so reviews are just recreational activity for me at the moment.)

snebel29:
I just merged master to solved the conflict with `asset/assets_vfsdata.go` then I regenerated to ensure all last changes were really codified into `assets_vfsdata.go`

All looks good except for `go.mod`, two indirect dependencies were added I'm still not fully sure where do they come from, do you know what's going on? everything seems to be working well though.

snebel29:
Hmmm, tests failed when I pushed the mentioned dependencies above, I guess those come from my environment, I just manually removed them.

snebel29:
Yes I had the same feeling, but I was unable to have some good looking just using bootstrap classes, I will keep this in mind in case eventually I find a more elegant way of doing so.

Issue #202, opened 2018-09-13
Reject inconsistent metrics at push time (and create a synthetic metric about it)


fvigotti:
and instead it should log the malformed data that has received and then discard it..
thre is already an issue in collectd exporter.. and I've already discussed it there  
https://github.com/prometheus/collectd_exporter/issues/56#issuecomment-419131935

I have no skill in go and zero time to help, sorry :(

fvigotti:
Hi @beorn7 , In understand that could be quite invasive , but having the whole application file to return metrics ( maybe some thousands of metrics ) because one of hundreds endpoint sent an invalid one it's a wrong behaviour, if you or other pushgateway developer are worried about performance or being too invasive change, let's user enable it with a arg-flag...
I don't know the internals.. but O(n²)  seems **a lot** to me.. ther eare no way to check validity of incoming data which is cheaper ?

beorn7:
First of all, if you have thousands of metrics on your PGW, the odds are you are using it wrong. See https://prometheus.io/docs/practices/pushing/ .

The PGW is fairly light-weight. In general, I would recommend to us one per team or service needing it. In that way, collisions of metrics are less likely, and a failure won't affect larger groups.

About the O(n²): This is just in the nature of the problem. Checking if the total exposition the PGW would offer to scrape is consistent is broadly proportional to the number of metrics pushed. Also, the check has to be performed on each push. Thus, if you double the number of pushers, you double the number of checks, and each check takes twice as long as double the metrics are pushed. Computational effort is now increased by a factor of four. One might improve this with some hash tables tracking the already pushed metrics, but it makes the code even more involved.

In any case, a rejected push needed to be represented in the scrape in _some_ way. You must be able to alert on it.

beorn7:
#94 is related but more concerned with the way how to handle the case if inconsistent metrics have made it into the Pushgateway. If this issue results in not letting this happen in the first place, #94 would be obsolete.

rombie:
Is there any update on this ? I tried with the latest version available now (0.9.1) but get the same set of errors. I do not understand the logic at all as to why /metrics get has to fail... it is the push that should fail, not get. Other option is to provide a get end-point. If so, then clients can call get before attempting the push... (I still don't udnerstand what O(n^2) referred to here..). Please do not fail GET on /metrics. Scraping will stop working due to this issue (and it was working few months ago)

beorn7:
The state is that the problem is well understood and a solution is sketched out in my head.
[...]

Issue #94, opened 2016-09-19
Mark in the web UI if metrics are inconsistent, and consider better ways to represent it in /metrics


tcolgate:
The javascript prometheus client library incorrectly has the process_cpu_seconds_total registered as a gauge (due to wanting to set it to a specific value, and the counter type not supporting that).

Pushing this metric to the push gateway causes the push gateway to crash. Pushgateway's /metrics returns

`expected counter in metric process_cpu_seconds_total label:<name:"instance" value:"tristans:8181" > label:<name:"job" value:"tristan-test" > gauge:<value:0.128 >`

This is also the case with metrics other than those used by the pushgateway itself, e.g., getting two jobs to post different values for a test metr, one posting gauge, and the other posting counter causes pushgateway's /metrics to return:

`expected counter in metric metric_test label:<name:"instance" value:"tristans:8182" > label:<name:"job" value:"tristan-test" > gauge:<value:2 >`

This should probably either be handled, or rejected more gracefully when the erroneous metric is posted.

tcolgate:
just to clairfy, there's no process panic, /metrics gives an error (we health check on /metrics on the push gateways, so our container manager kills it off)

beorn7:
The problem with checking on posting is that the inconsistency only manifests itself upon assembling the complete /metrics scrape (smelling like O(n²) ). Thus, we needed to do so upon each push to the Pushgateway.

The current architecture tries to accept pushes really fast and processes them later, so it would be hard to return errors to the pusher. Also, it's hard to find out the "true culprit" — and only one of the two pushers inconsistent with each other will be notified.

Thus, the current behavior was introduced deliberately. But I'm happy to hear ideas how to do it better.

tcolgate:
Could the type be ignored for subsequent pushes after the first? (clearly this means the scrape will advertise a different type then what was pushed). I guess this is mostly frustrating becase, at present, the types have no practical value.
  Could the web interface warn about the collision?

tcolgate:
yes, I can appreciate there's not really a good solution. I guess that
"ideally" the entire name:type would be the "primary key", and you could
happily serve both. But I think that breaks a great many assumptions
elsewhere.

I've got fixes locally for the prom-client side of this, so it's much less
of a concern. This does seem like a relatively easy "accidental denial of
service", waiting to happen. But then people aren't hitting it regularly,
so it may well be a non-issue.

beorn7:
[...]

PR #290, opened 2019-09-19
Check pushed metrics immediately and reject them if inconsistent

beorn7:
This is finally a solution to #202 (and as a byproduct to #94).

The original wisdom used to be that checking pushed metrics during the
push should not be done because it is inherently an O(n\*m) problem: n
is the number of metrics in the PGW, and m is the number of pushes in
a certain time-frame. Upon each push, a check has to run, and that
check has a cost linear with the total number of metrics. Thus, if you
scale up something by a factor of 2, you can expect twice as many
pushers pushing in total twice as many metrics, in twice as many push
operations, resulting in 2\*2=4 times the cost of checking
metrics. Since we didn't want to let pushers wait for a potentially
expensive check, we just accepted the push, as long as it was
parseable, and deferred the error detection to scrapes.

The result was that a single inconsistent push would result in an
error when scraping the PGW. The bright side here is that the problem
will be spotted quite easily, and there is a strong incentive to fix
it. The problematic part is that a single rogue pusher is affecting
all systems that depend on the same PGW, possibly at inconvenient
times.

In sane use cases for the PGW, scrapes are happening more often than
pushes. Let's say a database backup pushes its success or failure
outcome once per day. Scrapes happen a couple of times per
minute. While the whole O(n*m) calculation above is not wrong, the
reality is that the scrapes will inflict the cost anyway. If the check
is too expensive per push, it will be even more so per scrape. Also,
pushers shouldn't be too sensitive to slightly enlarged push
times. Finally, if a PGW is so heavily loaded with metrics that the
check actually takes prohibitively long, we can be quite sure that
this PGW is abused for something it was not meant for (like turning
Prometheus into a push-based system).

The solution is conceptually quite easy: On each push, first feed the
push into a copy of the current metrics state (which has a certain
cost but it is anyway done whenever somebody looks at the web UI) and
then simulate a scrape. In that way, we can reuse the validation power
of prometheus/client_golang. We don't even need to rearchitecture the
whole framework of queueing pushes and deletions. We only need to talk
back via an optional channel. Pushes will now get a 200 or a 400
rathen than the "(almost) always 202" response we had before.

Easy, isn't it?

Unfortunately, the easy change became _way_ more invasive than I
anticipated. Here is an overview of the changes in this commit (also
see the changes in README.md, which explain the usage side of things):

- Previously, the PUT request was executed in two parts: First, delete
  all metrics in the group, then add the newly pushed metrics to the
  group. That's now a problem. If the 2nd part fails, it will leave
  behind an empty group rather than the state before the whole PUT
  request, as the user will certainly expect. Thus, the operation had
  to be made atomic. To accomplish that, the `storage.WriteRequest`
  now has an additional field `Replace` to mark replacement of the
  whole group if true. This field is in addition to the back channel
  mentioned above (called `Done` in the `WriteRequest`).

- To enable alerting on failed pushes, there is now a new metric
  `push_failure_time_seconds` with the timestamp of the last failed
  push (or 0 if there never was a failed push). Ideally, we would
  rename the existing metric `push_time_seconds` to
  `push_success_time_seconds`, but that would break all existing uses
  of that metric. Thus, I left the name as is, although
  `push_time_seconds` is only updated upon a successful push. (This
  allows alerting on `push_failure_time_seconds > push_time_seconds`
  and tracks at the same time the last successful push.) There are
  some subtle aspects of implementing this correctly, mostly around
  the problem that a successful push should leave a pre-existing
  failure timestamp intact and vice versa.

- Previously, the push handler code already had a bunch of validation
  and sanitation code (block timestamps, sanitize labels) and also
  added the `push_time_seconds` metric. Now that the storage layer is
  doing the consistency check for the pushed metric, it is way cleaner
  to have all the other validation and sanitation code there as
  well. (Arguably, it was already wrong to have too much logic in the
  push handler.) Also, the decision needed to properly set the
  `push_time_seconds` and `push_faiure_time_seconds` metrics can now
  only be made after the consistency check, so those metrics have to
  be set in the storage layer anyway. This change of responsibility
  also changed the tests quite a bit. The push handler tests now test
  only that the right method calls happen to the storage, while the
  many tests for validation, consistency checks, adding the push
  timestamp metrics, and proper sanitation are now part of the storage
  tests. (Digression: One might argue that the cleanest solution is a
  separate validation layer in front of a dumb storage layer. Since
  the only implementation of the storage for now is the disk storage
  (and a mock storage for testing), I decided to not make that split
  prematurely.)

- Logging of rejected pushes happens ot error level now. Strictly
  speaking, it is just a user error and not an error of the PGW
  itself, if it is happening. But experience tells us that users first
  check the logs, and they usualyl don't run on debug or even info
  level. I hope this will avoid a whole lot of user support questions.
  Groups where the last push has failed are also marked in the web UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Examples:
 
         curl -X DELETE http://pushgateway.example.org:9091/metrics/job/some_job
         
-* Delete all metrics in all groups (requires to enable the admin api`--web.enable-admin-api`):
+* Delete all metrics in all groups (requires to enable the admin API via the command line flag `--web.enable-admin-api`):
 
         curl -X PUT http://pushgateway.example.org:9091/api/v1/admin/wipe
 
@@ -202,10 +202,12 @@ timestamps.
 If you think you need to push a timestamp, please see [When To Use The
 Pushgateway](https://prometheus.io/docs/practices/pushing/).
 
-In order to make it easier to alert on pushers that have not run recently, the
-Pushgateway will add in a metric `push_time_seconds` with the Unix timestamp
-of the last `POST`/`PUT` to each group. This will override any pushed metric by
-that name.
+In order to make it easier to alert on failed pushers or those that have not
+run recently, the Pushgateway will add in the metrics `push_time_seconds` and
+`push_failure_time_seconds` with the Unix timestamp of the last successful and
+failed `POST`/`PUT` to each group. This will override any pushed metric by that
+name. A value of zero for either metric implies that the group has never seen a
+successful or failed `POST`/`PUT`.
 
 ## API
 
@@ -277,20 +279,24 @@ header. (Use the value `application/vnd.google.protobuf;
 proto=io.prometheus.client.MetricFamily; encoding=delimited` for protocol
 buffers, otherwise the text format is tried as a fall-back.)
 
-The response code upon success is always 202 (even if the same
-grouping key has never been used before, i.e. there is no feedback to
-the client if the push has replaced an existing group of metrics or
-created a new one).
+The response code upon success is either 200 or 400. A 200 response implies a
+successful push, either replacing an existing group of metrics or creating a
+new one. A 400 response can happen if the request is malformed or if the pushed
+metrics are inconsistent with metrics pushed to other groups or collide with
+metrics of the Pushgateway itself. An explanation is returned in the body of
+the response and logged on error level.
+
+In rare cases, it is possible that the Pushgateway ends up with an inconsistent
+set of metrics already pushed. In that case, new pushes are also rejected as
+inconsistent even if the culprit is metrics that were pushed earlier. Delete
+the offending metrics to get out of that situation.
 
 _If using the protobuf format, do not send duplicate MetricFamily
 proto messages (i.e. more than one with the same name) in one push, as
 they will overwrite each other._
 
-A successfully finished request means that the pushed metrics are
-queued for an update of the storage. Scraping the push gateway may
-still yield the old results until the queued update is
-processed. Neither is there a guarantee that the pushed metrics are
-persisted to disk. (A server crash may cause data loss. Or the push
+Note that the Pushgateway doesn't provide any strong guarantees that the pushed
+metrics are persisted to disk. (A server crash may cause data loss. Or the push
 gateway is configured to not persist to disk at all.)
 
 A `PUT` request with an empty body effectively deletes all metrics with the
@@ -351,12 +357,14 @@ The default port the Pushgateway is listening to is 9091. The path looks like:
 The Pushgateway exposes the following metrics via the configured
 `--web.telemetry-path` (default: `/metrics`):
 - The pushed metrics.
-- For each pushed group, a metric `push_time_seconds` as explained above.
+- For each pushed group, a metric `push_time_seconds` and
+  `push_failure_time_seconds` as explained above.
 - The usual metrics provided by the [Prometheus Go client library](https://github.com/prometheus/client_golang), i.e.:
   - `process_...`
   - `go_...`
   - `promhttp_metric_handler_requests_...`
-- A number of metrics specific to the Pushgateway, as documented by the example scrape below.
+- A number of metrics specific to the Pushgateway, as documented by the example
+  scrape below.
 
 ```
 # HELP pushgateway_build_info A metric with a constant '1' value labeled by version, revision, branch, and goversion from which pushgateway was built.
@@ -385,7 +393,23 @@ pushgateway_http_requests_total{code="202",handler="push",method="post"} 6
 pushgateway_http_requests_total{code="400",handler="push",method="post"} 2
 
 ```
-  
+
+### Alerting on failed pushes
+
+It is in general a good idea to alert on `push_time_seconds` being much farther
+behind than expected. This will catch both failed pushes as well as pushers
+being down completely.
+
+To detect failed pushes much earlier, alert on `push_failure_time_seconds >
+push_time_seconds`.
+
+Pushes can also fail because they are malformed. In this case, they never reach
+any metric group and therefore won't set any `push_failure_time_seconds`
+metrics. Those pushes are still counted as
+`pushgateway_http_requests_total{code="400",handler="push"}`. You can alert on
+the `rate` of this metric, but you have to inspect the logs to identify the
+offending pusher.
+
 ## Development
 
 The normal binary embeds the web files in the `resources` directory.
--- a/handler/push.go
+++ b/handler/push.go
@@ -19,14 +19,12 @@ import (
 	"io"
 	"mime"
 	"net/http"
-	"sort"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/golang/protobuf/proto"
 	"github.com/julienschmidt/httprouter"
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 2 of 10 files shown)


beorn7:
@grobie you are still my dedicated PGW reviewer. :o)

@brian-brazil please sanity-check the `push_time_seconds` vs. `push_failure_time_seconds` metrics and the resulting alerting strategy (see README.md).

@snebel29 as you are getting deeper into the Pushgateway codebase, you might fancy helping reviewing this.

storage/diskmetricstore.go
```diff
@@ -345,3 +453,110 @@ func extractPredefinedHelpStrings(g prometheus.Gatherer) (map[string]string, err
...
+	return newTimestampGauge(pushMetricName, pushMetricHelp, groupingLabels, t)
+}
+
+func newPushFailedTimestampGauge(groupingLabels map[string]string, t time.Time) *dto.MetricFamily {
+	return newTimestampGauge(pushFailedMetricName, pushFailedMetricHelp, groupingLabels, t)
+}
+
+func newTimestampGauge(name, help string, groupingLabels map[string]string, t time.Time) *dto.MetricFamily {
+	var ts float64
+	if !t.IsZero() {
```
brian-brazil: Why do you need this if?

beorn7:
> Does the write channel still make sense, given we have to wait for the write to complete? It would be simpler to make that synchronous.

Perhaps. But the channel has the advantage of nice queueing behavior under contention. Also, in cases where you don't need the error feedback (or you are sure that an error cannot happen, like the delete case), you can simply leave out the feedback channel and still enjoy the old behavior.

README.md
```diff
@@ -277,20 +279,24 @@ header. (Use the value `application/vnd.google.protobuf;
...
+The response code upon success is either 200 or 400. A 200 response implies a
+successful push, either replacing an existing group of metrics or creating a
+new one. A 400 response can happen if the request is malformed or if the pushed
+metrics are inconsistent with metrics pushed to other groups or collide with
+metrics of the Pushgateway itself. An explanation is returned in the body of
+the response and logged on error level.
+
+In rare cases, it is possible that the Pushgateway ends up with an inconsistent
+set of metrics already pushed. In that case, new pushes are also rejected as
+inconsistent even if the culprit is metrics pushed earlier. Delete the
```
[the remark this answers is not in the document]
beorn7: Fixed.

storage/diskmetricstore.go
```diff
@@ -32,9 +34,15 @@ import (
 )
 
 const (
-	writeQueueCapacity = 1000
+	pushMetricName       = "push_time_seconds"
+	pushMetricHelp       = "Last Unix time when this group was changed in the Pushgateway."
```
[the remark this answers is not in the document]
beorn7: Yeah, I guess we have to change that. Will create inconsistent help string warnings, but they will go away eventually…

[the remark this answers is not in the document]
beorn7: My thought was that the old help string is fine because a failed push will not really _change_ the group. It will just leave it unchanged (with the exception of the failure timestamp metric). But I guess, a clearer wording is better.

storage/diskmetricstore.go
beorn7: Because the zero value of time.Time is not unixtime 0. Unixtime zero is 1970-01-01 00:00:00 UTC, Go time.Time zero value is 0001-01-01 00:00 UTC).

beorn7:
Another point for keeping the current `SubmitWriteRequest`: It is called at three locations in non-test code. Two of them don't need the feedback channel, only one does.

brian-brazil:
:+1:

beorn7:
Thanks, @brian-brazil .
I'll merge this now, but @grobie 's and @snebel29 's input is still welcome.
I won't cut a release before #287 is merged anyway.

snebel29:
Sorry I was out there without internet, I'm back now again. I will happy try to review future PR :+1:

handler/push.go
```diff
@@ -119,20 +108,28 @@ func Push(
...
 		ms.SubmitWriteRequest(storage.WriteRequest{
 			Labels:         labels,
 			Timestamp:      now,
 			MetricFamilies: metricFamilies,
+			Replace:        replace,
+			Done:           errCh,
 		})
-		w.WriteHeader(http.StatusAccepted)
+		for err := range errCh {
+			http.Error(
```
grobie: The documentation of [`http.Error`](https://golang.org/pkg/net/http/#Error) states "... the caller should ensure no further writes are done to w.", but it seems you're calling Error multiple times here, and therefore also write to `w` multiple times.

beorn7: Correct. I doubt that with the current code, we will ever see more than one error, but it's in principle possible (and intended to be possible) to send more than one error te tho channel. Even that will just trigger a warning from `http.Error`, but it's still something to avoid. I'll create a small follow-up PR.

jpdstan:
> check has to run, and that check has a cost linear with the total number of metrics

Is there a way we could speed this check up? From the docs it sounds like the only thing this check does is:

> Metrics of the same name must have the same type, even if they are pushed to different groups, and there must be no duplicates, i.e. metrics with the same name and the exact same label pairs. Pushes that would lead to inconsistencies are rejected with status code 400.

this sounds like something we could optimize with a hash-based data structure no?

beorn7:
Happy to review PRs making the check more efficient. (But note that this is probably more involved than it looks.)

beorn7:
In more detail: This PR simply utilizes the scrape-time check prometheus/client_golang performs to ensure the scraped exposition is valid. This check happens from scratch and has a cost linear to the number of exposed metrics. This has been done under the assumption that the PGW is scraped much more often than metrics are pushed to it, so optimizing the push path is not a priority.

To do a constant-time push check, you needed to recreate the whole check logic of prometheus/client_golang and keep it in sync.

If you push more often than you scrape, you are probably using the PGW for something it's not built for. That's why I'm reluctant to introduce a lot of hard-to-maintain and prone-to-introduce-new-bugs code for that unintended use case.

Having said that, if a PR demonstrates code that is simple enough so that it's not a big deal, I wouldn't oppose it.

jpdstan:
Thanks for the details! Can see why this is non-trivial.

> If you push more often than you scrape

This isn't the case for us, but our pushes are just getting bigger and bigger as our users add more metrics. We're just getting to a point that Pushgateway seems to be hitting practical limits, wherein it can't keep up with its write queue.

We're running Prometheus (Cortex) and PGW at a fairly large scale and don't really have a choice for some applications but to use PGW. At this point we're thinking of deploying multiple, smaller instances, rather than our central PGW today. Was hoping that a more optimized consistency check would buy us some more headroom until then. 

If you had any suggestions on how to run PGW at a higher scale, please let us know! So far the only other recommended alternative we've seen is to use the Node textfile collector, but that won't work for our K8s applications afaict.

xkilian:
You coul start sending data using Prometheus remote write, or instead if of pushing to push gateway send to redis. We wrote a simple scraper for redis for this purpose.
The push gateway is severely limited when compared to redis.
Cheers

beorn7:
The PGW is not designed to turn Prometheus into a push-based monitoring system. Scalability is just one aspect of it. There are many conceptual drawbacks.

If you are already using Cortex, why not run small Prometheus servers locally that remote-write to Cortex rather than running local jobs that push to PGW (which is then scraped by Prometheus wich then remote-writes to Cortex)?

The upcoming Prometheus agent might help as a more lightweight alternative to running a fully fledged Prometheus. (Or, in the meantime, you could look at https://github.com/grafana/agent , parts of which are the inspiration for the Prometheus agent.)

But this gets us really off-topic. A closed PR is anyway not a good place for a discussion, and if it is about ways of avoiding PGW, it's also not really about the PGW anymore. https://prometheus.io/community/ lists better channels for this kind of discussion.

PR #291, opened 2019-09-23
Avoid multiple calls of `http.Error` in push handler

beorn7:


beorn7:
I thought about that, too. But it appeared a lot of plumbing for a case that doesn't even happen yet.
(And if it happens, perhaps the amount of errors is so excessive that we don't even want to return a giant body.)
I'll add a TODO.

PR #292, opened 2019-09-24
Feat/enable shutdown via http request

snebel29:
I still have to find where to update documentation, in the mean while how does this implementation looks like?

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,22 +354,22 @@ The default port the Pushgateway is listening to is 9091. The path looks like:
         
 ## Management API
 
-Pushgateway provides a set of management API to ease automation and integrations.
+The Pushgateway provides a set of management API to ease automation and integrations.
 
 * Available endpoints:
  
 | HTTP_METHOD |  PATH | DESCRIPTION |
 | :-------: | :-----| :----- |
-| GET    | /-/healthy |  Return 200 whenever the Pushgateway is healthy. |
-| GET    | /-/ready |  Return 200 whenever the Pushgateway is ready. |
+| GET    | /-/healthy |  Returns 200 whenever the Pushgateway is healthy. |
+| GET    | /-/ready |  Returns 200 whenever the Pushgateway is ready to serve traffic. |
 
-* The following endpoints are disabled by default and can be enabled via the `--web.enable-lifecycle` flag.
+* The following endpoint is disabled by default and can be enabled via the `--web.enable-lifecycle` flag.
 
 | HTTP_METHOD |  PATH | DESCRIPTION |
 | :-------: | :-----| :----- |
 | PUT    | /-/quit |  Triggers a graceful shutdown of Pushgateway. |
 
-An alternative way trigger a graceful shutdown is by sending a `SIGTERM` to the Pushgateway process.
+Alternatively, a graceful shutdown can be triggered by sending a `SIGTERM` to the Pushgateway process.
 
 ## Exposed metrics
 
--- a/README.md
+++ b/README.md
@@ -352,6 +352,25 @@ The default port the Pushgateway is listening to is 9091. The path looks like:
 
         curl -X PUT http://pushgateway.example.org:9091/api/v1/admin/wipe
         
+## Management API
+
+Pushgateway provides a set of management API to ease automation and integrations.
+
+* Available endpoints:
+ 
+| HTTP_METHOD |  PATH | DESCRIPTION |
+| :-------: | :-----| :----- |
+| GET    | /-/healthy |  Return 200 whenever the Pushgateway is healthy. |
+| GET    | /-/ready |  Return 200 whenever the Pushgateway is ready. |
+
+* The following endpoints are disabled by default and can be enabled via the `--web.enable-lifecycle` flag.
+
+| HTTP_METHOD |  PATH | DESCRIPTION |
+| :-------: | :-----| :----- |
+| PUT    | /-/quit |  Triggers a graceful shutdown of Pushgateway. |
+
+An alternative way trigger a graceful shutdown is by sending a `SIGTERM` to the Pushgateway process.
+
 ## Exposed metrics
 
 The Pushgateway exposes the following metrics via the configured
--- a/main.go
+++ b/main.go
@@ -145,7 +145,32 @@ func main() {
 		level.Error(logger).Log("err", err)
 		os.Exit(1)
 	}
-	go interruptHandler(l, logger)
+
+	quitCh := make(chan struct{})
+	quitHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "Requesting termination... Goodbye!")
+		close(quitCh)
+	})
+
+	forbiddenAPINotEnabled := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusForbidden)
+		w.Write([]byte("Lifecycle APIs are not enabled"))
+	})
+
+	if *enableLifeCycle {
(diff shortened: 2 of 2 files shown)


beorn7:
User documentation is really only in the README.md for now.

beorn7:
Will have a look ASAP.

snebel29:
In a side note, I didn't dig much on it but I've noticed that running `make` add a couple of indirect dependencies to `go.mod` and extra stuff to `modules.txt`

beorn7:
'make' is also exercising the Go module workflow (to check if `go.mod` is tidy and such). Depending on your go version, things might change slightly. Or perhaps `go.mod` is in fact not tidy at the moment. (Those are just theories that come to my head spontaneously.)

README.md
```diff
@@ -352,6 +352,25 @@ The default port the Pushgateway is listening to is 9091. The path looks like:
 
         curl -X PUT http://pushgateway.example.org:9091/api/v1/admin/wipe
         
+## Management API
+
+Pushgateway provides a set of management API to ease automation and integrations.
```
beorn7: Consistency nit: We usually say "The Pushgateway …" in this README.

README.md
```diff
@@ -352,6 +352,25 @@ The default port the Pushgateway is listening to is 9091. The path looks like:
...
+
+* Available endpoints:
+ 
+| HTTP_METHOD |  PATH | DESCRIPTION |
+| :-------: | :-----| :----- |
+| GET    | /-/healthy |  Return 200 whenever the Pushgateway is healthy. |
+| GET    | /-/ready |  Return 200 whenever the Pushgateway is ready. |
```
beorn7: Perhaps "… is ready to serve traffic."?

beorn7: I would use 3rd person, i.e. "Returns" instead of "Return" for both lines above.

README.md
```diff
@@ -352,6 +352,25 @@ The default port the Pushgateway is listening to is 9091. The path looks like:
...
+
+* The following endpoints are disabled by default and can be enabled via the `--web.enable-lifecycle` flag.
```
beorn7: For now, it's only one endpoint, so "The following endpoint is disabled…"

README.md
```diff
@@ -352,6 +352,25 @@ The default port the Pushgateway is listening to is 9091. The path looks like:
...
+
+| HTTP_METHOD |  PATH | DESCRIPTION |
+| :-------: | :-----| :----- |
+| PUT    | /-/quit |  Triggers a graceful shutdown of Pushgateway. |
+
+An alternative way trigger a graceful shutdown is by sending a `SIGTERM` to the Pushgateway process.
```
beorn7: Grammar is a bit off here. How about a simpler formulation like: "Alternatively, a graceful shutdown can be triggered by sending a `SIGTERM` to the Pushgateway process."

main.go
```diff
@@ -144,7 +145,32 @@ func main() {
...
+
+	quitCh := make(chan struct{})
+	quitHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintf(w, "Requesting termination... Goodbye!")
+		close(quitCh)
+	})
+
+	forbiddenAPINotEnabled := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusForbidden)
+		w.Write([]byte("Lifecycle APIs are not enabled"))
```
beorn7: "Lifecycle API is not enabled."

main.go
```diff
@@ -144,7 +145,32 @@ func main() {
...
+		r.Handler("PUT", *routePrefix+"/-/quit", quitHandler)
+		r.Handler("POST", *routePrefix+"/-/quit", quitHandler)
+	} else {
+		r.Handler("PUT", *routePrefix+"/-/quit", forbiddenAPINotEnabled)
+		r.Handler("POST", *routePrefix+"/-/quit", forbiddenAPINotEnabled)
+	}
+
+	r.Handler("GET", "/-/quit", http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		w.Write([]byte("Only POST or PUT requests allowed"))
```
beorn7: Nit: End sentence with a period.

main.go
```diff
@@ -190,10 +216,17 @@ func computeRoutePrefix(prefix string, externalURL *url.URL) string {
...
 	signal.Notify(notifier, os.Interrupt, syscall.SIGTERM)
-	<-notifier
-	level.Info(logger).Log("msg", "received SIGINT/SIGTERM; exiting gracefully...")
+
+	select {
+	case <-notifier:
+		level.Info(logger).Log("msg", "received SIGINT/SIGTERM; exiting gracefully...")
+		break
+	case <-quitCh:
+		level.Warn(logger).Log("msg", "Received termination request via web service, exiting gracefully...")
```
beorn7: Generally, we don't capitalize log messages (even if they are almost sentences), i.e. use "received" and not "Received".

main.go
```diff
@@ -190,10 +216,17 @@ func computeRoutePrefix(prefix string, externalURL *url.URL) string {
 	return prefix
 }
 
-func interruptHandler(l net.Listener, logger log.Logger) {
+func exitHandler(l net.Listener, quitCh <-chan struct{}, logger log.Logger) {
```
beorn7: `Handler` is kind of confusing as a name because this is not an HTTP handler as above. It was already confusing prior to this PR, but one could have argued that "signal handling" is common Unix lingo. Now that it is handling more than a signal, that argument doesn't hold anymore. How about `closeListenerOnQuit` or something, and add a doc comment, because the function is doing two things now, which is harder to see at a first glance. Something like "closeListenerOnQuite closes the provided listener upon closing the provided quitCh or upon receiving a SIGTERM."

README.md
snebel29: Actually you're right, I copied it from the Prometheus documentation I think I will fix the text there as well.

https://prometheus.io/docs/prometheus/latest/management_api/#quit

main.go
snebel29: Yes, I will fix the message [here](https://github.com/prometheus/prometheus/blob/27bbc235e08ba00161909fa19a47e6420d6d99e7/web/web.go#L358) too

main.go
snebel29: I had similar feelings about this one initially :smile:

snebel29: I just added SIGINT signal to your proposed comment... I even start ending my sentences into github comments with periods :smile: .

snebel29:
I think this should be good now.

snebel29:
I just came to a realization about log messages being capitalized, from the [contribution guidelines](
https://github.com/prometheus/prometheus/blob/master/CONTRIBUTING.md) several documents are referred, one of them being [go wiki](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) and excluding log messages from not being capitalized, in Prometheus code base you can actually find both cases (capitalized and non capitalized) commonly

https://github.com/prometheus/prometheus/blob/08c55c119f39093e18b2bb9cba5c5619dc4ea0e1/cmd/prometheus/main.go#L296

https://github.com/prometheus/prometheus/blob/08c55c119f39093e18b2bb9cba5c5619dc4ea0e1/cmd/prometheus/main.go#L693

I'm happy to stick to whatever is decided (so far non capitalized?), and even to enforce it in the prometheus ecosystem code base with a PRs.

beorn7:
Yeah, consistency is hard. Some of us are nitpicky enough and try to enforce it at least in new code. It's not the end of the world if a log message is capitalized now and then. Not sure if it is worth to go through the codebase and fix it all (but be my guest).

Issue #288, opened 2019-09-13
Handle or avoid hash collisions


beorn7:
It always has been a known issue, kind of: We hash the grouping key, using a fairly weak hash function (fnv64a, https://github.com/prometheus/common/blob/master/model/signature.go#L33). And we do not handle hash collisions. No incident reported so far, but I think the current behavior is not good enough for v1. Plus, it's not really required to do the hashing at all. We are not in a hot codepath or something. It was only done out of convenience back then.

The downside: The hash value makes it all the way down to the disk format. Which means we have to change/convert/stay-compatible. Another reason to do this before v1.

PR #293, opened 2019-10-03
Avoid hash collisions in the storage by not hashing grouping labels

beorn7:
Instead, simply put into a string what was hashed before and use that
string as the key in the map instead of the uint64 hash value used before.

Hashing was essentially done for convenience, ignoring the ill effect
of a hash collision, should it ever occur. In contrast to other uses
of hashing of label maps, the hashing isn't needed here for
performance reasons. Neither CPU cycles nor memory usage is critical
here.

The downside is that the on-disk format changes. This commit includes
code to convert the old format if encountered. This code is planned to
be removed prior to the v1 release of the PGW.

Note about testing the conversion: I have verified manually that it
works and was planning to add a test. However, that test would require
adding a fixture in the old format. As this code will be removed in
the next released, I decided it's not worth the
complication. (Automated tests are there to make sure things keep
working after code changes. In this case, the next change will be to
remove the code to be tested.)

### DIFF
--- a/storage/diskmetricstore.go
+++ b/storage/diskmetricstore.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -60,6 +61,10 @@ type mfStat struct {
 	pos    int  // Where in the result slice is the MetricFamily?
 	copied bool // Has the MetricFamily already been copied?
 }
+
+// legacyGroupingKeyToMetricGroup is like GroupingKeyToMetricGroup but uses the
+// old uint64 grouping key (a hash value) instead of a string grouping key.
+type legacyGroupingKeyToMetricGroup map[uint64]MetricGroup
 
 // NewDiskMetricStore returns a DiskMetricStore ready to use. To cleanly shut it
 // down and free resources, the Shutdown() method has to be called.
@@ -264,7 +269,7 @@ func (dms *DiskMetricStore) processWriteRequest(wr WriteRequest) {
 	dms.lock.Lock()
 	defer dms.lock.Unlock()
 
-	key := model.LabelsToSignature(wr.Labels)
+	key := groupingKeyFor(wr.Labels)
 
 	if wr.MetricFamilies == nil {
 		// No MetricFamilies means delete request. Delete the whole
@@ -307,7 +312,7 @@ func (dms *DiskMetricStore) setPushFailedTimestamp(wr WriteRequest) {
 	dms.lock.Lock()
 	defer dms.lock.Unlock()
 
-	key := model.LabelsToSignature(wr.Labels)
+	key := groupingKeyFor(wr.Labels)
 
 	group, ok := dms.metricGroups[key]
 	if !ok {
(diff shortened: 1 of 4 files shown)


beorn7:
This fixes #288 .

@snebel29 perhaps you are interested in reviewing this. Also nominated @grobie as the designated PGW reviewer. :o)

storage/diskmetricstore.go
```diff
@@ -61,6 +62,10 @@ type mfStat struct {
 	copied bool // Has the MetricFamily already been copied?
 }
 
+// legacyGroupingKeyToMetricGroup is like GroupingKeyToMetricGroup but uses the
+// old uint64 grouping key (a hash value) instead of a string grouping key.
+type legacyGroupingKeyToMetricGroup map[uint64]MetricGroup
```
snebel29: So you keep `legacyGroupingKeyToMetricGroup` type in this file instead of in `storage/interface.go` together `GroupingKeyToMetricGroup` type to make it easier to remove it (because is all together) later on? what's the reasoning behind?

storage/diskmetricstore.go
```diff
@@ -425,7 +430,25 @@ func (dms *DiskMetricStore) restore() error {
...
+	}
+	d = gob.NewDecoder(f)
+	if err := d.Decode(&legacyMetricGroups); err != nil {
+		return err // That's a real failure then, unrelated to the format change.
+	}
+	for _, mg := range legacyMetricGroups {
+		dms.metricGroups[groupingKeyFor(mg.Labels)] = mg
+	}
+	level.Info(dms.logger).Log("msg", "conversion from v0.9 legacy format successful, next persisting will happen in current format")
+	return nil
```
snebel29: So you don't encapsulate this code into a restore legacy function because will anyway be removed soon? (together with v1 release)

storage/diskmetricstore.go
beorn7: I don't want to export it. It's just a private type to handle the conversion. An unexported type in a file called `interface.go` with otherwise all the exported types would be weird.

storage/diskmetricstore.go
beorn7: Kind of. I guess it could be extracted as a separate function. But with passing in an open file handle and such, it would be a quite complex call. I think the inline code is just easier to understand. The fact that we'll remove the code very soon is only important as it removes the code maintenance argument.

storage/diskmetricstore.go
```diff
@@ -437,6 +460,31 @@ func copyMetricFamily(mf *dto.MetricFamily) *dto.MetricFamily {
 	}
 }
 
+// groupingKeyFor creates a grouping key from the provided map of grouping
```
cstyan: Usually I think it would be nice to include a comment for what the format of the type is since we don't have a `type groupingKey struct` but in this case the format is pretty clear from the function implementation and the tests. Up to you if you want to clarify the format or uniqueness of the key for a given set of label names/values here.

storage/diskmetricstore.go
```diff
@@ -437,6 +460,31 @@ func copyMetricFamily(mf *dto.MetricFamily) *dto.MetricFamily {
...
+		return ""
+	}
+
+	labelNames := make([]string, 0, len(labels))
+	for labelName := range labels {
+		labelNames = append(labelNames, labelName)
+	}
+	sort.Strings(labelNames)
+
+	sb := strings.Builder{}
```
cstyan: TIL about `strings.Builder`

beorn7: It's fairly recent, introduced in Go 1.10.

storage/diskmetricstore.go
beorn7: I added an explanation how the grouping key is created and that it is unique.

Issue #299, opened 2019-10-16
"Metric families have inconsistent help strings" with default push_time_seconds metric


kfdm:
## Bug Report
**What did you do?**

After upgrading pushgateway to 0.10.0, existing metrics threw errors on `push_time_seconds`

To create a simple test case, we use the following code, slightly modified from the Pushgateway example

```shell
# Push some metrics to pushgateway
cat <<EOF | curl --verbose --data-binary @- http://localhost:9091/metrics/job/some_job/instance/some_instance
# TYPE some_metric counter
some_metric{label="val1"} 42
# TYPE another_metric gauge
# HELP another_metric Just an example.
another_metric 2398.283
EOF

# Trigger a scrape to verify things work
curl http://localhost:9091/metrics > /dev/null

# Push some more metrics to push gateway
cat <<EOF | curl --verbose --data-binary @- http://localhost:9091/metrics/job/some_job/instance/another_instance
# TYPE something_else counter
something_else{label="val1"} 42
EOF

# Trigger another scrape for testing
curl http://localhost:9091/metrics > /dev/null
```

We then launch push gateway, and run the above script in a second window. It works as expected and close (Ctrl+C) when we're done testing

```
./pushgateway-0.9.1.darwin-amd64/pushgateway --persistence.file=test --log.level=debug 
level=info ts=2019-10-16T02:53:48.453Z caller=main.go:79 msg="starting pushgateway" version="(version=0.9.1, branch=HEAD, revision=b4410a3a361a770a084007ef50aebf42d0e8793c)"
level=info ts=2019-10-16T02:53:48.454Z caller=main.go:80 build_context="(go=go1.12.7, user=root@556a5d403380, date=20190801-14:04:02)"
level=debug ts=2019-10-16T02:53:48.454Z caller=main.go:81 msg="external URL" url=
level=debug ts=2019-10-16T02:53:48.454Z caller=main.go:82 msg="path prefix used externally" path=
level=debug ts=2019-10-16T02:53:48.454Z caller=main.go:83 msg="path prefix for internal routing" path=
level=error ts=2019-10-16T02:53:48.454Z caller=diskmetricstore.go:86 msg="could not load persisted metrics" err="gob: decoding into local type *storage.GroupingKeyToMetricGroup, received remote type GroupingKeyToMetricGroup"
level=info ts=2019-10-16T02:53:48.455Z caller=main.go:133 listen_address=:9091
[...]

PR #300, opened 2019-10-16
Add documentation about inconsistent metrics

beorn7:
In particular, explain the issue with the help string for
`push_time_seconds` in v0.9 and earlier.

@kfdm what do you think? This should address #299.


kfdm:
Looks good to me. My primary reason for filing a bug, was just so there was some reference to link to. Putting it in the README seems like a good plan :)

beorn7:
Yeah, and also, that inconsistency story was never really explained nicely.

PR #303, opened 2019-10-17
Enforce static conversion for common values

sylr:
I've a pretty busy PG which `/` endpoints delivers 120MB of HTML data and takes forever to render.

Although pagination would be the correct answer to the problem, I'm wondering if using static conversion from float64 to string for most common values wouldn't save some CPU while rendering.

It's a page taken from the kube-state-metrics optimization book:

https://github.com/kubernetes/kube-state-metrics/blob/v1.8.0/pkg/metric/metric.go#L106-L109

Signed-off-by: Sylvain Rabot <s.rabot@lectra.com>

### DIFF
--- a/handler/status.go
+++ b/handler/status.go
@@ -67,6 +67,11 @@ func Status(
 			t := template.New("status")
 			t.Funcs(template.FuncMap{
 				"value": func(f float64) string {
+					if f == 0 {
+						return "0"
+					} else if f == 1 {
+						return "1"
+					}
 					return strconv.FormatFloat(f, 'f', -1, 64)
 				},
 				"timeFormat": func(t time.Time) string {


handler/status.go
```diff
@@ -67,6 +67,11 @@ func Status(
 			t := template.New("status")
 			t.Funcs(template.FuncMap{
 				"value": func(f float64) string {
+					if f == 0 {
+						return "0"
+					} else if f == 1 {
```
beorn7: You don't need the `else` here.

sylr:
So I tried to build the PGW against a version of client_model with added GetStringValue() functions on metric structs but it shown no CPU usage improvements.

https://github.com/prometheus/pushgateway/compare/v0.9.1..sylr:d413687467772d9ea8397fae12ba6a8e932d3acc?expand=1

I'll close this as it is less optimized that what I tested.

PR #314, opened 2019-12-20
Update dependencies and cut v1.0.1

beorn7:
This is mostly to push actually working non-amd64 container images.


beorn7:
@brian-brazil as you have taken care of the other two release PRs, perhaps you'd like this one, too.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.0.1 / 2019-12-21
+
+[ENHANCEMENT] Remove excessive whitespace from HTML templates. #302
```
brian-brazil: Is this even worth a mention? It's not really user visible.

beorn7: I was on the fence here, but since the cognitive load of a two-line changelog is still quite limited, I thought I mention it, just in case somebody is confused that suddenly the response size drops on the status page, or somebody was annoyed before about all the empty lines and tabs when inspecting the HTML code and has now a reason to rejoice over the holidays.

Issue #313, opened 2019-12-19
Introduce flag to disable push checks


benamazing:
Hi, 

We use pushgateway to collect app metrics. We used version 0.7 pushgateway before, and it worked well. Today we upgraded to v1.0, found the push performance decreased a lot, the avg push duration increased from less than 1 seconds to 300+ seconds. Is this a bug or any logic changed in pushgateway version > v0.7.0 ?

![image](https://user-images.githubusercontent.com/7868153/71170607-a2345980-2296-11ea-8383-e534232caf0a.png)

beorn7:
This might very well be the check of pushed metrics introduced in [v0.10](https://github.com/prometheus/pushgateway/releases/tag/v0.10.0). May I ask how many metrics live on your Pushgateway? How long does it usually take to _scrape_ your Pushgateway? (The check was introduced with the rationale that pushes, with the check, will not take longer than scrapes, and that in sane setups, pushes are less frequent than scrapes.)

benamazing:
> This might very well be the check of pushed metrics introduced in [v0.10](https://github.com/prometheus/pushgateway/releases/tag/v0.10.0). May I ask how many metrics live on your Pushgateway? How long does it usually take to _scrape_ your Pushgateway? (The check was introduced with the rationale that pushes, with the check, will not take longer than scrapes, and that in sane setups, pushes are less frequent than scrapes.)

We have aound 100 clients push the metrics to the same pushgateway. Each clients would push around 15,000 metrics with the interval of 60 seconds.

And prometheus scrapes pushgateway every 60 seconds as well. Each scrape cost around 6-8 seconds.

beorn7:
Wow, that's wild… and indeed nothing the PGW was ever designed for.

6–8 seconds scrape is long but probably borderline OK. Same is true for the duration of a push. However, I can imagine what is happening here: With 100 clients each pushing every 60s, you have a push every 600ms. If the first push takes as long as a scrape, there will be more than 10 pushes coming in while it is still ongoing. Or in other words: Many pushes will overlap, slowing down each individual push, making even more pushes overlap. Death spiral ensues…

And yes, you push _way_ more frequently than you scrape.

With the load and usage pattern you have, it looks you are using the PGW to turn Prometheus into a push-based monitoring system. That's an explicit anti-goal of the PGW. I don't think the PGW should be optimized to cater for those use-cases. But perhaps we should consider a protection against too many concurrent pushes and reject further pushes if too many are still ongoing.

[...]

PR #318, opened 2020-01-06
Add flag to disable push-time consistency checks

beorn7:
@benamazing @belm0 please have a look. This should solve your problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ All pushes are done via HTTP. The interface is vaguely REST-like.
 
 ### URL
 
-The default port the push gateway is listening to is 9091. The path looks like
+The default port the Pushgateway is listening to is 9091. The path looks like
 
     /metrics/job/<JOB_NAME>{/<LABEL_NAME>/<LABEL_VALUE>}
 
@@ -337,8 +337,8 @@ proto messages (i.e. more than one with the same name) in one push, as
 they will overwrite each other._
 
 Note that the Pushgateway doesn't provide any strong guarantees that the pushed
-metrics are persisted to disk. (A server crash may cause data loss. Or the push
-gateway is configured to not persist to disk at all.)
+metrics are persisted to disk. (A server crash may cause data loss. Or the
+Pushgateway is configured to not persist to disk at all.)
 
 A `PUT` request with an empty body effectively deletes all metrics with the
 specified grouping key. However, in contrast to the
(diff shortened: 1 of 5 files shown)


beorn7:
Fixes #313.

README.md
```diff
@@ -305,12 +317,15 @@ header. (Use the value `application/vnd.google.protobuf;
 proto=io.prometheus.client.MetricFamily; encoding=delimited` for protocol
 buffers, otherwise the text format is tried as a fall-back.)
 
-The response code upon success is either 200 or 400. A 200 response implies a
-successful push, either replacing an existing group of metrics or creating a
-new one. A 400 response can happen if the request is malformed or if the pushed
-metrics are inconsistent with metrics pushed to other groups or collide with
-metrics of the Pushgateway itself. An explanation is returned in the body of
-the response and logged on error level.
+The response code upon success is either 200, 202, or 400. A 200 response
```
belm0: I wonder if 202 will catch existing custom clients off guard (naively expecting 200).

Returning 200 isn't incorrect, so it's a question of value and whether any client would actually need to discern this (I doubt it).

handler/handler_test.go
```diff
@@ -101,9 +101,9 @@ func TestHealthyReady(t *testing.T) {
 func TestPush(t *testing.T) {
 	mms := MockMetricStore{}
 	mmsWithErr := MockMetricStore{err: errors.New("testerror")}
-	handler := Push(&mms, false, false, logger)
-	handlerWithErr := Push(&mmsWithErr, false, false, logger)
-	handlerBase64 := Push(&mms, false, true, logger)
+	handler := Push(&mms, false, true, false, logger)
+	handlerWithErr := Push(&mmsWithErr, false, true, false, logger)
+	handlerBase64 := Push(&mms, false, true, true, logger)
```
belm0: A comment listing the bool arg names would help readability.

README.md
beorn7: 202 is what we had before the push checks got introduced. So all clients, custom or not, that worked with PGW v0.9 and earlier should also work with this. 

My assumption here is therefore that all the existing clients are fine with it, and we get the differentiation of the two different modes for free.

handler/handler_test.go
beorn7: Good point. Added.

beorn7:
@belm0 I assume I have sufficiently addressed your points. Merging now.

PR #325, opened 2020-01-27
gitattributes: use wildcard

simon04:
Follow-up for https://github.com/prometheus/pushgateway/pull/323

Without the wildcard, resources/static/bootstrap-4.3.1-dist/js/bootstrap.bundle.js is still considered …


simon04:
Awesome, thank you. Now the language statistics make sense:

Screenshot_2020-01-29 prometheus pushgateway

PR #329, opened 2020-02-19
Change httprouter to common/route

drumilpatel2000:

### DIFF
--- a/handler/delete.go
+++ b/handler/delete.go
@@ -21,8 +21,6 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/route"
 
 	"github.com/prometheus/pushgateway/storage"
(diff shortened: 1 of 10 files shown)


handler/push.go
```diff
@@ -207,3 +207,12 @@ func splitLabels(labels string) (map[string]string, error) {
 	}
 	return result, nil
 }
+
+// TODO: remove this, Don't know the reason but seems that route.Params is Not working
+// Subtitute for route.Params
+func urlParams(ctx context.Context, p string) string {
```
drumilpatel2000: @beorn7 Don't Know the reason but it seems that `route.Params` is not working in this situation. Add manual function that does the same thing.

drumilpatel2000: It is correct I just implemented this in the wrong way

handler/handler_test.go
```diff
@@ -76,6 +75,16 @@ func (m *MockMetricStore) Ready() error {
 	return nil
 }
 
+func ctxWithParams(params map[string]string, mainReq *http.Request) context.Context {
```
drumilpatel2000: @beorn7 Probably changing this function can solve problems occurring in test cases

handler/delete.go
```diff
@@ -14,14 +14,15 @@
 package handler
 
 import (
+	"context"
 	"fmt"
+	"github.com/prometheus/common/route"
```
beorn7: Nit: this should go into the next block (after line 27).

handler/handler_test.go
```diff
@@ -76,6 +75,16 @@ func (m *MockMetricStore) Ready() error {
...
+	ctx := mainReq.Context()
+
+	for key, value := range params {
+		ctx = context.WithValue(ctx, key, value)
```
beorn7: Isn't it more appropriate to use `route.WithParam` for this?

Perhaps that also fixes the problem you are seeing in the tests.

handler/push.go
```diff
@@ -162,9 +163,10 @@ func Push(
 				handler,
 			)))
 
-	return func(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		fmt.Println(r.Context())
 		mtx.Lock()
-		ps = params
+		ctx = r.Context()
```
beorn7: I think you don't need this hack anymore. It was only needed for the httprouter.Params as they wouldn't play with the promhttp instrumentation.

The whole wrapping function can probably go. Locking and context retrieval can happen up there in the `handler` function.

beorn7:
NB: I haven't done a thorough review on this yet. I was mostly focused on finding the reason for the test failures and about removing the hacks we needed to use the httprouter package.

You might be able to use the `WithInstrumentation` method from the `route` package here, too (to also solve instrumentation for the API endpoints, see TODO entry).

beorn7:
:tada: for passing tests.

Is this ready for the final review?

drumilpatel2000:
Yes ready for the real checks(final reviews) :wink:

main.go
```diff
@@ -192,8 +225,7 @@ func main() {
 		apiPath = *routePrefix + apiPath
 	}
 
-	// TODO: Instrument API endpoints.
-	av1 := route.New()
+	av1 := route.New().WithInstrumentation(instrumentHandler(apiPath + "/v1/"))
```
beorn7: I think the trailing `/` in `"/v1/"` has to go.

main.go
```diff
@@ -46,8 +45,42 @@ import (
...
+		},
+		[]string{"handler", "method"},
+	)
+	responseSize = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "pushgateway_http_response_size_bytes",
+			Help:    "Histogram of response size for HTTP requests.",
+			Buckets: prometheus.ExponentialBuckets(100, 10, 7),
+		},
+		[]string{"handler", "method"},
```
beorn7: These are good metrics, but I think they are overkill for the Pushgateway. Or let's say I would only introduce this level of detail if somebody asks for it.

Also, these metrics compete now with the metrics in handler/metrics.go.

Let's for now only use the metrics we already have. I'd suggest to leave `httpPushDuration` and `httpPushDuration` exactly as they are, with the explicit handling in the push handler, and only use `httpCnt` in a generic way.

main.go
```diff
@@ -46,8 +45,42 @@ import (
...
 func init() {
 	prometheus.MustRegister(version.NewCollector("pushgateway"))
+	prometheus.MustRegister(requestDuration)
+	prometheus.MustRegister(responseSize)
+}
+
+func instrumentHandler(prefix string) func(handlerName string, handler http.HandlerFunc) http.HandlerFunc {
+	return func(handlerName string, handler http.HandlerFunc) http.HandlerFunc {
+		handlerLabel := prometheus.Labels{"handler": prefix + handlerName}
+		return promhttp.InstrumentHandlerDuration(
```
beorn7: As said above, I'd only do the instrumentation with `httpCnt` from `handler/metrics.go` for now. Please try to simplify the handler label (in line 75 above) so that we end up with the same label values as in the existing instrumentation (for the handlers that are already instrumented), i.e. `delete`, `wipe`, `status`, `healthy`, `ready`, `static`, `push` (and the new ones in the same style). With the code as is, you get fairly long and detailed labels that aren't necessarily helpful (e.g. `handler="/metrics/job/:job/*labels"` instead of just `handler="push"`). I know that the common/route package doesn't make that easy as it insists on using the path for `handlerName`, but at the very least we should not break current users by changing the existing handler label. Sorry for that…

main.go
```diff
@@ -46,8 +48,34 @@ import (
...
+		[]string{"handler", "code", "method"},
+	)
+)
+
 func init() {
 	prometheus.MustRegister(version.NewCollector("pushgateway"))
+	prometheus.MustRegister(httpCnt)
+}
+
+func getFunctionName(i interface{}) string {
```
drumilpatel2000: Don't Know this should be used or not but it works fine if implemented. Made httpCnt common for all routes as used directly by route instrumentation. @beorn7 Any thoughts

main.go
drumilpatel2000: Quite confused. How I can implement this feature but for a temporary solution I am manual adding it.

drumilpatel2000: @beorn7 Any advice

beorn7: I think I have an idea. I'll show you in a separate PR. It's too hard to explain in the comments. And I would like to do it based on an older commit of yours.

beorn7:
My idea didn't work out as I thought, but as a byproduct of me trying to implement it, I did the instrumentation. I have created a new PR #329, which includes most of the commits from here. Let's continue there.

PR #331, opened 2020-02-25
Continuation of #329

beorn7:
Hi @drumilpatel2000 , I thought my idea would work, but in the end it was way more tedious than I had imagined. In the end, I just instrumented explicitly, probably what you wanted to do anyway.

Nevertheless, now that I have done it, let's take it from here. I also fixed a few nits from the review of #329, and I noticed that `/api/v1/wipe` wasn't working anymore (which I fixed here).

### DIFF
--- a/handler/delete.go
+++ b/handler/delete.go
@@ -35,7 +35,7 @@ func Delete(ms storage.MetricStore, jobBase64Encoded bool, logger log.Logger) fu
 	var mtx sync.Mutex // Protects ps.
 
 	instrumentedHandler := promhttp.InstrumentHandlerCounter(
-		httpCnt.MustCurryWith(prometheus.Labels{"handler": "delete"}),
+		HTTPCnt.MustCurryWith(prometheus.Labels{"handler": "delete"}),
 		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			job := route.Param(r.Context(), "job")
 			if jobBase64Encoded {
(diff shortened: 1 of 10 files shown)


beorn7:
OK, @drumilpatel2000 , now you are the reviewer. Let me know what's missing. (o:

handler/delete.go
```diff
@@ -21,33 +21,32 @@ import (
...
-func Delete(ms storage.MetricStore, jobBase64Encoded bool, logger log.Logger) func(http.ResponseWriter, *http.Request, httprouter.Params) {
-	var ps httprouter.Params
+func Delete(ms storage.MetricStore, jobBase64Encoded bool, logger log.Logger) func(http.ResponseWriter, *http.Request) {
 	var mtx sync.Mutex // Protects ps.
 
 	instrumentedHandler := promhttp.InstrumentHandlerCounter(
-		httpCnt.MustCurryWith(prometheus.Labels{"handler": "delete"}),
-		http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
-			job := ps.ByName("job")
+		HTTPCnt.MustCurryWith(prometheus.Labels{"handler": "delete"}),
```
drumilpatel2000: This can be removed by using similar wrap function used in api/v1/api.go

beorn7: Indeed. And if we have a helper function for that here in handler, we can also use it in the `wrap` for the API and don't need to export HTTPCnt. Just a minute…

beorn7:
This should work now. PHAL.

beorn7:
There we go…

PR #332, opened 2020-02-27
Add test for APIs

### LINKED ISSUE #184: Expose metrics metadata in HTTP API

drumilpatel2000:

Fixes #184

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,2 +1,0 @@
-# Default ignored files
-/workspace.xml
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,2 @@
+# Default ignored files
+/workspace.xml
--- a/api/v1/api_test.go
+++ b/api/v1/api_test.go
@@ -15,13 +15,16 @@ package v1
 import (
 	"bytes"
 	"encoding/json"
+	"math"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 	"time"
 
-	"github.com/go-kit/kit/log"
+	dto "github.com/prometheus/client_model/go"
 
+	"github.com/go-kit/kit/log"
+	"github.com/golang/protobuf/proto"
 	"github.com/prometheus/pushgateway/storage"
 )
 
@@ -37,7 +40,106 @@ var (
 		"build2": "value2",
 		"build3": "value3",
 	}
+
+	mf1 = &dto.MetricFamily{
+		Name: proto.String("mf1"),
+		Type: dto.MetricType_UNTYPED.Enum(),
+		Metric: []*dto.Metric{
+			{
+				Label: []*dto.LabelPair{
+					{
+						Name:  proto.String("instance"),
+						Value: proto.String("instance2"),
+					},
+					{
+						Name:  proto.String("job"),
+						Value: proto.String("job1"),
+					},
+				},
+				Untyped: &dto.Untyped{
+					Value: proto.Float64(-3e3),
+				},
+			},
+		},
+	}
+
+	mf2 = &dto.MetricFamily{
+		Name: proto.String("mf2"),
+		Help: proto.String("doc string 2"),
+		Type: dto.MetricType_GAUGE.Enum(),
+		Metric: []*dto.Metric{
+			{
+				Label: []*dto.LabelPair{
+					{
+						Name:  proto.String("basename"),
+						Value: proto.String("basevalue2"),
+					},
+					{
+						Name:  proto.String("instance"),
+						Value: proto.String("instance2"),
+					},
+					{
+						Name:  proto.String("job"),
+						Value: proto.String("job1"),
+					},
+					{
+						Name:  proto.String("labelname"),
+						Value: proto.String("val2"),
+					},
+				},
+				Gauge: &dto.Gauge{
+					Value: proto.Float64(math.Inf(+1)),
+				},
+			},
+			{
+				Label: []*dto.LabelPair{
+					{
+						Name:  proto.String("instance"),
+						Value: proto.String("instance2"),
+					},
+					{
+						Name:  proto.String("job"),
+						Value: proto.String("job1"),
+					},
+					{
+						Name:  proto.String("labelname"),
+						Value: proto.String("val1"),
+					},
+				},
+				Gauge: &dto.Gauge{
+					Value: proto.Float64(math.Inf(-1)),
+				},
+			},
+		},
+	}
+
+	grouping1 = map[string]string{
+		"job":      "job1",
+		"instance": "instance1",
+	}
 )
+
+func templateMetrics(metricStore storage.MetricStore) []interface{} {
+	familyMaps := metricStore.GetMetricFamiliesMap()
+	res := make([]interface{}, 0)
+	for _, v := range familyMaps {
+		metricResponse := make(map[string]interface{})
+		metricResponse["labels"] = v.Labels
+		metricResponse["last_push_successful"] = v.LastPushSuccess()
+		for name, metricValues := range v.Metrics {
+			metricFamily := metricValues.GetMetricFamily()
+			uniqueMetrics := metrics{
+				Type:      metricFamily.GetType().String(),
+				Help:      metricFamily.GetHelp(),
+				Timestamp: metricValues.Timestamp,
+				Metrics:   makeEncodableMetrics(metricFamily.GetMetric(), metricFamily.GetType()),
+			}
+			metricResponse[name] = uniqueMetrics
+		}
+		res = append(res, metricResponse)
+	}
+	return res
+}
 
 func compareMaps(mainMap map[string]interface{}, testMap map[string]string) bool {
 	if len(mainMap) != len(testMap) {
@@ -46,6 +148,18 @@ func compareMaps(mainMap map[string]interface{}, testMap map[string]string) bool
 
 	for key, _ := range mainMap {
 		if mainMap[key].(string) != testMap[key] {
+			return false
+		}
+	}
+	return true
+}
+
+func compareArray(a, b []byte) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i, v := range a {
+		if v != b[i] {
 			return false
 		}
 	}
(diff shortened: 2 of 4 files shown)


.idea/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Default ignored files
+/workspace.xml
```
beorn7: I guess you have accidentally added this file.

api/v1/api_test.go
```diff
@@ -0,0 +1,196 @@
...
+	if !compareMaps(responseFlagData, testFlags) {
+		t.Errorf("Wanted following flags %v, got %v.", testFlags, responseFlagData)
+	}
+
+	if !compareMaps(responseBuildInfo, testBuildInfo) {
+		t.Errorf("Wanted following build info %v, got %v.", testBuildInfo, responseBuildInfo)
+	}
+}
+
+func TestMetricsAPI(t *testing.T) {
```
drumilpatel2000: @beorn7 I implemented TestMetricsApi but don't know how vigorous testing should be done
Can you help me with the same

api/v1/api_test.go
```diff
@@ -82,3 +114,83 @@ func TestStatusAPI(t *testing.T) {
...
+	req, err := http.NewRequest("GET", "http://example.org/", &bytes.Buffer{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	w := httptest.NewRecorder()
+
+	testResponse := response{}
+	TestApi.metrics(w, req)
+	json.Unmarshal(w.Body.Bytes(), &testResponse)
```
beorn7: Wouldn't it be easier to provide a golden raw JSON string and compare it directly rather than unmarshalling and comparing the pieces?

beorn7:
@drumilpatel2000 are you still working on this? Do you need more help?

drumilpatel2000:
> @drumilpatel2000 are you still working on this? Do you need more help?

Definitely, I am working on it but due to midterms in my college I was quite busy. Will resume work by 6th.

beorn7:
Thank you very much. And take your time.

drumilpatel2000:
Free from midterms. Ready to rock on :fireworks:

api/v1/api_test.go
```diff
@@ -82,3 +196,76 @@ func TestStatusAPI(t *testing.T) {
...
+	requiredResponse, _ := json.Marshal(&response{
+		Status: statusSuccess,
+		Data:   templateMetrics(dms),
+	})
+
+	if expected, got := http.StatusOK, w.Code; expected != got {
+		t.Errorf("Wanted status code %v, got %v.", expected, got)
+	}
+
+	if !compareArray(requiredResponse, w.Body.Bytes()) {
```
drumilpatel2000: @beorn7 How about this way to compare responses? :smile:

drumilpatel2000:
@beorn7 No idea why are test failing in Travis but the same test in circle ci is just fine.

api/v1/api_test.go
```diff
@@ -0,0 +1,273 @@
...
+		t.Errorf("Wanted response %v, got %v.", string(requiredResponse), w.Body.String())
+	}
+}
+
+func metricFamiliesMap(mfs ...*dto.MetricFamily) map[string]*dto.MetricFamily {
```
beorn7: I'd also copy the doc comment from where you copied this (diskmetricstore_test.go) because it is otherwise confusing to the reader why you are doing this.

Having said that, as this gets close to a significant amount of not entirely trivial code, perhaps we should move it to a `testutil` package (or perhaps even `internal/testutil`, see https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface ).

api/v1/api_test.go
```diff
@@ -0,0 +1,273 @@
...
+	json.Unmarshal(w.Body.Bytes(), &testResponse)
+	jsonData := testResponse.Data.(map[string]interface{})
+	responseFlagData := jsonData["flags"].(map[string]interface{})
+	responseBuildInfo := jsonData["build_information"].(map[string]interface{})
+
+	if expected, got := http.StatusOK, w.Code; expected != got {
+		t.Errorf("Wanted status code %v, got %v.", expected, got)
+	}
+
+	if !compareMaps(responseFlagData, testFlags) {
```
beorn7: Would `reflect.DeepEqual` work here?

api/v1/api_test.go
```diff
@@ -0,0 +1,273 @@
...
+		t.Errorf("Wanted following flags %v, got %v.", testFlags, responseFlagData)
+	}
+
+	if !compareMaps(responseBuildInfo, testBuildInfo) {
```
beorn7: Same here.

api/v1/api_test.go
```diff
@@ -0,0 +1,273 @@
...
+		metricResponse := make(map[string]interface{})
+		metricResponse["labels"] = v.Labels
+		metricResponse["last_push_successful"] = v.LastPushSuccess()
+		for name, metricValues := range v.Metrics {
+			metricFamily := metricValues.GetMetricFamily()
+			uniqueMetrics := metrics{
+				Type:      metricFamily.GetType().String(),
+				Help:      metricFamily.GetHelp(),
+				Timestamp: metricValues.Timestamp,
+				Metrics:   makeEncodableMetrics(metricFamily.GetMetric(), metricFamily.GetType()),
```
beorn7: The problem with this approach is that you are calling `makeEncodableMetrics` to create the expected test result. You are not really testing that part, you are essentially creating a self-fulfilling prophecy: Calling the same function in the test creates the same results.

That's why I would go for an approach where you just hardcode a "golden" JSON string and compare it directly.

api/v1/api_test.go
```diff
@@ -0,0 +1,273 @@
...
+					{
+						Name:  proto.String("instance"),
+						Value: proto.String("instance2"),
+					},
+					{
+						Name:  proto.String("job"),
+						Value: proto.String("job1"),
+					},
+				},
+				Untyped: &dto.Untyped{
```
beorn7: I'd use at least one of the more complex metric types, if not both, i.e. Summary and/or Histogram.

api/v1/api_test.go
```diff
@@ -0,0 +1,273 @@
...
+
+	for err := range errCh {
+		t.Fatal("Unexpected error:", err)
+	}
+
+	w = httptest.NewRecorder()
+
+	TestApi.metrics(w, req)
+
+	requiredResponse, _ = json.Marshal(&response{
```
beorn7: As for the other, simpler example above, I'd hardcode the expected JSON string here and then compare to `w.Body.String()`. (There maybe parts in the JSON response that's not static, like timestamps. You could either leave them out (e.g. via a regexp match rather than a direct string comparison) or fill them in, making the hardcoded string a bit less hardcoded.)

As described below, the problem is that some code paths are now the same in the test and in the code to be tested, so that they will come up with the same result even if the outcome is not what we want.

api/v1/api_test.go
beorn7: As commented separately, having a Histogram and or Summary in it, would be good. Since it is reflected in other comments, I'll resolve this one.

api/v1/api_test.go
```diff
@@ -0,0 +1,273 @@
...
+		Metric: []*dto.Metric{
+			{
+				Label: []*dto.LabelPair{
+					{
+						Name:  proto.String("instance"),
+						Value: proto.String("instance2"),
+					},
+					{
+						Name:  proto.String("job"),
+						Value: proto.String("job1"),
```
beorn7: I'd put a few potentially problematic characters here, like different quotes and multy-byte UTF-8 characters, e.g. `Björn'"`

beorn7:
Hehe, #333 fixed Travis, indeed.

api/v1/api_test.go
drumilpatel2000: Probably reflect.DeepEqual was not giving correct answer

testutil/metric_families.go
```diff
@@ -0,0 +1,27 @@
+package testutil
```
beorn7: License header is missing in this file.

api/v1/api_test.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	if !compareMaps(responseBuildInfo, testBuildInfo) {
+		t.Errorf("Wanted following build info %q, got %q.", testBuildInfo, responseBuildInfo)
+	}
+}
+
+func TestMetricsAPI(t *testing.T) {
+	dms := storage.NewDiskMetricStore("", 100*time.Millisecond, nil, logger)
+	TestApi := New(logger, dms, testFlags, testBuildInfo)
```
beorn7: Should be `testAPI` according to general Go coding style.

api/v1/api_test.go
```diff
@@ -0,0 +1,225 @@
...
+					SampleCount: proto.Uint64(0),
+					SampleSum:   proto.Float64(0),
+				},
+			},
+		},
+	}
+
+	grouping1 = map[string]string{
+		"job":      "Björn",
+		"instance": "instance1",
```
beorn7: We still need to test escaping, so how about turning this into
```Go
	grouping1 = map[string]string{
		"job":      "Björn",
		"instance": `inst'a"n\ce1`,
	}
```
?

api/v1/api_test.go
beorn7: `reflect.DeepEqual` fails because the two maps have a different type.

Rather than `compareMaps`, how about defining a function like
```Go
func convertMap(m map[string]string) map[string]interface{} {
	result := map[string]interface{}{}
	for k, v := range m {
		result[k] = v
	}
	return result
}
```

This will then allow to write this line as
```Go
if !reflect.DeepEqual(responseFlagData, convertMap(testFlags)) {
```

It's in sum not much less code than your approach, but I suspect that many readers will ask themselves why you didn't use `reflect.DeepEqual` and suspect a tricky reason. This makes it clearer that it is really just a plain comparison.

api/v1/api_test.go
beorn7: See above.

api/v1/api_test.go
```diff
@@ -0,0 +1,221 @@
...
+	result := map[string]interface{}{}
+	for k, v := range m {
+		result[k] = v
+	}
+	return result
+}
+
+func TestStatusAPI(t *testing.T) {
+	dms := storage.NewDiskMetricStore("", 100*time.Millisecond, nil, logger)
+	TestApi := New(logger, dms, testFlags, testBuildInfo)
```
beorn7: This also needs to be changed to `testAPI`.

api/v1/api_test.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	if !reflect.DeepEqual(responseBuildInfo, convertMap(testBuildInfo)) {
+		t.Errorf("Wanted following build info %q, got %q.", testBuildInfo, responseBuildInfo)
+	}
+}
+
+func TestMetricsAPI(t *testing.T) {
+	dms := storage.NewDiskMetricStore("", 100*time.Millisecond, nil, logger)
+	testApi := New(logger, dms, testFlags, testBuildInfo)
```
beorn7: Almost. :o)

It should be `testAPI` not `testApi`.

drumilpatel2000:
@beorn7 Next Target Documentations.

beorn7:
Thanks 1M! Merging now.

beorn7:
>  Next Target Documentations.

I guess it could be merged with the existing documentation for the admin API.

PR #335, opened 2020-03-11
Cut v1.2.0

beorn7:
I believe the addition of the new query API is the only user-visible change.

All the internal changes about HTTP routing should not have any impact visible from the outside. Even the new metrics are only affecting the query API, IIRC.

@drumilpatel2000


beorn7:
That was quick. I thought @drumilpatel2000 is already asleep. (o:

I'll change the release date to today then.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.2.0 / 2020-03-11
+
+* [FEATURE] Add an HTTP API to query metadata and runtime information. #184
```
roidelapluie: Maybe `HTTP API endpoint`?

roidelapluie: (If you change the date anyway..)

beorn7: It's actually two endpoints. I think talking about an API in general makes more sense.

beorn7: Actually, it should contain "metric" somehow, let me change that.

PR #345, opened 2020-05-08
Fix docker hub link in readme

loivis:


README.md
```diff
@@ -62,7 +62,7 @@ persisted (so that they survive restarts of the Pushgateway).
 
 ### Using Docker
 
-You can deploy the Pushgateway using the [prom/pushgateway](https://registry.hub.docker.com/u/prom/pushgateway/) Docker image.
+You can deploy the Pushgateway using the [prom/pushgateway](https://registry.hub.docker.com/r/prom/pushgateway/) Docker image.
```
beorn7, line 65: It looks like `https://hub.docker.com/r/prom/pushgateway` is the preferred URL nowadays.

loivis: Exactly. Didn't notice there is difference in host.

Issue #344, opened 2020-05-08
Pushing empty label values returns 405 Method Not Allowed


matevz:
Using `client_golang`:
```
myPusher = myPusher.Grouping("firstlabel", "abc")
myPusher = myPusher.Grouping("secondlabel", "")
myPusher = myPusher.Grouping("thirdlabel", "def")
```

results in generating PUT HTTP request
```
http://mypushgateway:9091/metrics/job/myjob/firstlabel/abc/secondlabel//thirdlabel/def
```

and when sending it to push gateway, it responds `405 Method Not Allowed`.

I'm not sure whether that's a push gateway caveat, URL encoding problem or `client_golang` issue, but empty label values are perfectly legal in Prometheus and should work fine.

Using pushgateway 1.2.0 and client_golang 1.5.1.

beorn7:
I guess it's mostly an URL encoding problem. Will investigate…

beorn7:
Sooo, this is even a bit deeper:

If a double-slash (`//`) is encountered, the Go ServeMux creates an HTTP 301 redirect to the cleaned path. Proper HTTP clients then change the method to GET when they encounter a redirect because POST and PUT are not considered idempotent. That's where your 405 is coming from. We could fix that change of method by sending a 307 or 308, but that wouldn't help us here because then the Pushgateway will complain about the wrong number of components.

We could probably hack deeply into the HTTP routing code to somehow allow `//` to make it through without a redirect, but that will be tedious and dirty.

The use case for empty grouping label values is probably rather narrow, but as you say, it's formally valid and I could even dream up some scenarios where it makes a difference from just leaving out the label with the empty value altogether (same reason why the PGW adds an empty `instance` label by default).

Luckily, there is a work-around using base64 encoding. Your PUT works as intended with the following URL: `http://mypushgateway:9091/metrics/job/myjob/firstlabel/abc/secondlabel@base64/=/thirdlabel/def`

Now we only need to teach that to the client libraries…

I'll send a PR to update the docs here on the PGW side.

beorn7:
With the updated docs, this issue is handled from the PGW side. Fixes for client libraries are happening as I'm typing this.

PR #346, opened 2020-05-12
Document how to use empty label values in the grouping key

beorn7:
"Fixes" #344.

@matevz

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,10 +266,20 @@ _will be overwritten to match the labels defined by the URL path!_
 If `job` or any label name is suffixed with `@base64`, the following job name
 or label value is interpreted as a base64 encoded string according to [RFC
 4648, using the URL and filename safe
-alphabet](https://tools.ietf.org/html/rfc4648#section-5). (Padding is
-optional.) This is the only way of using job names or label values that contain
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -278,25 +288,34 @@ Examples:
...
+      /metrics/job/directory_cleaner/path@base64/L3Zhci90bXA
+      
   If you are not using a client library that handles the encoding for you, you
   can use encoding tools. For example, there is a command line tool `base64url`
   (Debian package `basez`), which you could combine with `curl` to push from
   the command line in the following way:
   
       echo 'some_metric{foo="bar"} 3.14' | curl --data-binary @- http://pushgateway.example.org:9091/metrics/job/directory_cleaner/path@base64/$(echo -n '/var/tmp' | base64url)
+
+* To use the grouping key `job="example",first_label="",second_label="foobar"`,
```
brian-brazil, line 305: It's not quite clear that the issue here is the empty label. I'd suggest something like "To use a grouping key with an empty label value such as ..."

beorn7: Makes sense, updated.

brian-brazil:
:+1:

PR #361, opened 2020-07-18
show RemoteAddr when push metrics parse error

ozhiwei:
when some metrics pushed, but parse text error, the debug log like this:

```
level=debug ts=2020-07-18T06:58:41.917Z caller=push.go:111 msg="failed to parse text" err="text format parsing error in line 1: label value \"sc} 1361199104\" contains unescaped new-line"
```

we can't know the instance which is the problem

so add the HTTP RemoteAddr, then
```
level=debug ts=2020-07-18T07:30:41.923Z caller=push.go:111 addr=172.16.3.112:42812 msg="failed to parse text" err="text format parsing error in line 1: label value \"sc} 1361199104\" contains unescaped new-line"
```

### DIFF
--- a/handler/push.go
+++ b/handler/push.go
@@ -108,7 +108,7 @@ func Push(
 		}
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusBadRequest)
-			level.Debug(logger).Log("msg", "failed to parse text", "err", err.Error())
+			level.Debug(logger).Log("msg", "failed to parse text", "source", r.RemoteAddr, "err", err.Error())
 			return
 		}
 		now := time.Now()


handler/push.go
```diff
@@ -108,7 +108,7 @@ func Push(
 		}
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusBadRequest)
-			level.Debug(logger).Log("msg", "failed to parse text", "err", err.Error())
+			level.Debug(logger).Log("addr", r.RemoteAddr, "msg", "failed to parse text", "err", err.Error())
```
beorn7, line 111: How about making this fully consistent with the logging in line 146, i.e. also add r.Method, and call the r.RemoteAddr entry `source` rather than `addr`?

ozhiwei: great!, let me check!

Issue #362, opened 2020-07-29
UI: Metrics text alignment issue on zoom


belm0:
## Bug Report
**What did you do?**
View Metrics UI with a zoom percentage in browser

**What did you expect to see?**
normal display metrics items

**What did you see instead? Under which circumstances?**
some metric items with long text are center-aligned rather than left-aligned

![Screen Shot 2020-07-29 at 11 44 56 AM](https://user-images.githubusercontent.com/1708631/88750913-7ca6f100-d191-11ea-9ae7-5ca329ae4282.png)

**Environment**

* Pushgateway version:

v1.2.0

chunmeng:
Is inserting style attribute to the specific buttons acceptable?

![image](https://user-images.githubusercontent.com/2604020/95012079-185c4e00-0668-11eb-97d9-6bcd512fce73.png)

beorn7:
Closing as the ad-hoc solution should fix the problem at hand. Of course, if some web expert has a cleaner solution, I'm all ers.

PR #369, opened 2020-10-04
fix: issue #362: UI: Metrics text alignment issue on zoom

chunmeng:
Add style="text-align:left" to button in metric-panel


chunmeng:
@beorn7

beorn7 (approved):
I guess it would be cleaner with an own class in the CSS or something, but good enough for now to stop the bleeding.

beorn7:
Thank you.

PR #390, opened 2021-03-05
Use relative paths for assets in template.html if no pathPrefix is set

simonfrey:
Loading the assets always from toplevel `/static` is not a good practice and breaks our setup (as we deploy the pushgateway with a reverse proxy to a subpath)

This PR defaults to relative paths which works for the status handler on `/` and `/status`

No behavior is changed for configurations with path prefix

### DIFF
--- a/handler/status.go
+++ b/handler/status.go
@@ -104,6 +104,10 @@ func Status(
 				"goVersion": version.GoVersion,
 			}
 
+			if pathPrefix == "" {
+				pathPrefix = "."
+			}
+
 			d := &data{
 				MetricGroups: ms.GetMetricFamiliesMap(),
 				BuildInfo:    buildInfo,


handler/status.go
```diff
@@ -104,6 +105,10 @@ func Status(
 				"goVersion": version.GoVersion,
 			}
 
+			if pathPrefix != "" && !strings.HasSuffix(pathPrefix, "/") {
+				pathPrefix = pathPrefix + "/"
+			}
+
```
beorn7, line 111: Would it also work to just do the below and not change `template.html` at all?
```Go
	if pathPrefix == "" {
		pathPrefix = "."
	}
```

beorn7:
To make the test pass, you have to run `make assets`. (But if you don't change `template.html`, as proposed above, you don't need to do that.)

beorn7:
I have a vague recollection that we tried relative paths first and then had to change it to absolute paths because _something_ didn't work. But I cannot find evidence what it was. It's probably six years ago or so…

If nobody else remembers, we can try what you propose here and see what happens.

handler/status.go
simonfrey: Oh that also should work. Nice catch! 
Will updated PR

beorn7 (approved):
OK, :crossed_fingers: . Let's see what breaks…

simonfrey:
Tried it on our system and it works without a problem so far

simonfrey:
Is there a schedule when this change will land in a release? Maybe 1.4.1 ? 

I can't deploy the head of this repo to our prod env, but need a stable (release) state

beorn7:
Nothing else showed up that would have motivated me to cut a release. But enough time has passed. I'll cut a v1.4.1 in the next days.

Thanks for reminding me.

beorn7:
In fact, #404 should definitely go in. And I'll try if I can find a quick fix for #401 before releasing.

PR #397, opened 2021-04-01
Update template.html

nhathoangvy:
Add css for badge info get error

### DIFF
--- a/resources/template.html
+++ b/resources/template.html
@@ -71,7 +71,7 @@ limitations under the License.
 					<span class="toggle-icon glyphicon glyphicon-collapse-down"></span>
 					{{- $metricGroup := .}}
 					{{- range $i, $ln := .SortedLabels}}
-					<span class="badge {{if eq $ln "job"}}badge-warning{{else if eq $ln "instance"}}badge-primary{{else}}badge-info{{end}}">{{$ln}}="{{index $metricGroup.Labels $ln}}"</span>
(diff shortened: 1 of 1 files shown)


beorn7:
Thanks for your contribution. Could you sign the DCO? See https://github.com/prometheus/pushgateway/pull/397/checks?check_run_id=2244593779

Also, could you explain the change? Bonus points for screenshot before/after.

nhathoangvy:
Hi @beorn7 !
I want to separate API success and fail UI .
So just that .

beorn7:
Those badges don't always look as expected. Could you attach a screenshot before and after?

nhathoangvy:
Screen Shot 2021-04-09 at 08 27 29
@beorn7 like this

beorn7:
Thank you very much. Only now I understand what this is about. (I should have seen that earlier, but the screenshot was helpful indeed.)

So, the current behavior treats the `instance` and the `job` label special because they are actually special in a technical sense (the `job` label is mandatory, and the `instance` label is auto-set to `""` if not specified explicitly). However, an `error` label or a `result` label and their label values are just conventions. For the PGW and for Prometheus, they aren't special at all. There isn't even any general practice or convention in the Prometheus ecosystem to have any of those labels, not to speak about specific values. Therefore, I don't think we should color the labels here in the PGW UI solely based on a weak heuristics like "`failed` in the label value" or "the label name is `error`".

beorn7:
Assuming the lack of follow-up means my line of argument above is sound, I'll close thise.

PR #404, opened 2021-05-18
Ensure requests are checked while draining

dgl:
Fix issue where pushes received when draining didn't get labels, resulting in pushgateway saving bad data to the persist file and then failing to serve /metrics on restart.

(Potentially it would also be possible for inconsistent data to be accepted here, but the more noticeable issue is the side-effect of `checkWriteRequest` calling `sanitizeLabels` doesn't happen, so no labels are added).

### DIFF
--- a/storage/diskmetricstore_test.go
+++ b/storage/diskmetricstore_test.go
@@ -551,7 +551,7 @@ var (
 			},
 		},
 	}
-	mfUnlaballed = &dto.MetricFamily{
+	mfUnlabelled = &dto.MetricFamily{
 		Name: proto.String("mf_unlabelled"),
 		Help: proto.String("Metric with no labels to check sanitizeLabels."),
 		Type: dto.MetricType_GAUGE.Enum(),
@@ -912,7 +912,7 @@ func TestAddDeletePersistRestore(t *testing.T) {
 	dms.SubmitWriteRequest(WriteRequest{
 		Labels:         grouping6,
 		Timestamp:      ts5,
-		MetricFamilies: testutil.MetricFamiliesMap(mfUnlaballed),
+		MetricFamilies: testutil.MetricFamiliesMap(mfUnlabelled),
 	})
 	if err := dms.Shutdown(); err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


beorn7:
404 – pull request not found

beorn7:
Bad jokes aside, I'll have a look ASAP.

storage/diskmetricstore_test.go
```diff
@@ -892,17 +905,42 @@ func TestAddDeletePersistRestore(t *testing.T) {
...
+		newPushTimestampGauge(grouping5, ts5).Metric[0],
+		newPushTimestampGauge(grouping6, ts5).Metric[0],
 	)
 	pushFailedTimestamp.Metric = append(
-		pushFailedTimestamp.Metric, newPushFailedTimestampGauge(grouping5, time.Time{}).Metric[0],
+		pushFailedTimestamp.Metric,
+		newPushFailedTimestampGauge(grouping5, time.Time{}).Metric[0],
+		newPushFailedTimestampGauge(grouping6, time.Time{}).Metric[0],
 	)
+	mfLaballed := proto.Clone(mfUnlabelled).(*dto.MetricFamily)
```
beorn7, line 930: mfLaballed → mfLabelled

storage/diskmetricstore_test.go
```diff
@@ -892,17 +905,42 @@ func TestAddDeletePersistRestore(t *testing.T) {
...
+	// sanitizeLabels should add these labels to the unlabelled metric
```
beorn7, line 931: Nit: Start with capital letter and end with a period.

beorn7 (changes requested):
Thanks for the catch and for fixing it. Just a few spelling nits.

storage/diskmetricstore_test.go
```diff
@@ -927,9 +927,9 @@ func TestAddDeletePersistRestore(t *testing.T) {
 		newPushFailedTimestampGauge(grouping5, time.Time{}).Metric[0],
 		newPushFailedTimestampGauge(grouping6, time.Time{}).Metric[0],
 	)
-	mfLaballed := proto.Clone(mfUnlabelled).(*dto.MetricFamily)
-	// sanitizeLabels should add these labels to the unlabelled metric
-	mfLaballed.Metric[0].Label = []*dto.LabelPair{
+	mfLabelled := proto.Clone(mfUnlabelled).(*dto.MetricFamily)
+	// SanitizeLabels should add these labels to the unlabelled metric.
```
beorn7, line 931: Actually, sanitizeLabels is an actual method name. So I complained too soon. But anyway, we don't have to get even nitpickier.

beorn7 (approved):
Thanks again.

PR #413, opened 2021-08-04
Replace go-kit/kit with go-kit/log, update dependencies

mrueg:
As it was done in https://github.com/prometheus/common/pull/304

@roidelapluie


beorn7 (approved):
Thank you very much.

PR #428, opened 2021-10-03
Log shutdown gracefully

### LINKED ISSUE #427: Avoid an error level warning upon graceful shutdown
### Discussed in https://github.com/prometheus/pushgateway/discussions/426

<div type='discussions-op-text'>

<sup>Originally posted by **jiangpengjul** September 30, 2021</sup>

image

image
</div>

mjip:
Closes #427

@beorn7


beorn7:
Thanks for your contribution.

Unfortunately, this doesn't work as intended. The way the graceful shotdown is implemented right now is to close the socket. After that, `web.Serve` returns with on error (something like `accept tcp [::]:9091: use of closed network connection`). 

If you try out and run the code with your PR, then hit CTRL-C, you will still see the error-level log message.

mjip:
Thanks for picking that up! I added a check on the error message- unfortunately, the maintainers of the `net` package decided against exporting the `errClosing` error for use of closed network connections, so I couldn't check the returned error type- https://github.com/golang/go/issues/4373 I also added a comment for context about this change.

beorn7:
Hmm, I think relying on a certain substring in the error message is very brittle.

I'm not really an expert when it comes to HTTP serving in Go, but I assume the real problem is that we bluntly call `l.Close()` here: https://github.com/prometheus/pushgateway/blob/079d8d899a0fc47c3b159df2b3a4ffc555170466/main.go#L257

I have now done a bit of research, and my impression is that we should store a pointer to the `http.Server` in a variable and give it to the `closeListenerOnQuit` goroutine. The server is created here: https://github.com/prometheus/pushgateway/blob/079d8d899a0fc47c3b159df2b3a4ffc555170466/main.go#L198

Then, instead of calling `l.Close()`, we could call [`http.Server.Shutdown`](https://pkg.go.dev/net/http#Server.Shutdown). Which has two advantages: First, the error returned is now the exported `http.ErrServerClosed`. Second, we got an actual graceful shutdown for free, so we can get rid of this weird code: https://github.com/prometheus/pushgateway/blob/079d8d899a0fc47c3b159df2b3a4ffc555170466/main.go#L200-L203

The reason why things were done in this weird way is that it is really old code. The [`http.Server.Shutdown`](https://pkg.go.dev/net/http#Server.Shutdown) method was only added in Go1.8, and this Pushgateway code predates that Go version.

Does that make any sense?

mjip:
Your suggestion made sense to me, and I prefer it as well. Let me know if this is what you had in mind.

main.go
```diff
@@ -194,13 +194,21 @@ func main() {
...
-	time.Sleep(time.Second)
+	server := &http.Server{
+		Addr:    *listenAddress,
+		Handler: mux,
+	}
+
+	go shutdownServerOnQuit(server, quitCh, logger)
+	err = web.Serve(l, server, *webConfig, logger)
+
+	// In the case of a shutdown, log gracefully
```
beorn7, line 205: Nit: End sentence with a period.

And, even more nitpicking, "log gracefully" might not really nail it. How about "In the case of a graceful shutdown, do not log the error."

beorn7 (changes requested):
Code looks perfect. Thank you very much for removing this old wart.

I just have a tiny bit of nitpicking about a doc comment.

PR #457, opened 2022-03-18
[doc] better cmd in readme

qingfeng777:
Signed-off-by: liushengjie <liushengjie@qcraft.ai>

after copy, the 'EOF' of 'cat <<' command don't work, echo maybe better.


beorn7:
I like the here document more. Could you explain why it doesn't work anymore "after copy"?

beorn7:
Closing after 4 weeks without response. Feel free to follow up, it will be noticed.

Issue #469, opened 2022-05-10
New Release Soon?


andrewkratochvil:
Is it possible to get a new release soon?  There are several CVEs with the current release (1.4.2) that are caused by old dependencies.  It appears that the prombot has updated the dependencies and would fix several of the CVEs.  Thank you!

beorn7:
Yes, a new release is imminent.

Note that there is no evidence that any of the CVEs is affecting the Pushgateway. It's a bit of a trade-off if that justifies a new release (removing bugs that are probably irrelevant and trading them for yet unknown bugs that might very well be relevant…). I will still cut a release, simply because the previous one is so long ago.

beorn7:
The release is done: https://github.com/prometheus/pushgateway/releases/tag/v1.4.3

As said, no known issues with the old release, but this has updated dependencies and the binaries are built with Go 1.18.

svc-cfs-acnt:
Thank you for the new release!  Updating the dependencies resolved all the findings.

PR #471, opened 2022-05-28
Release v1.4.3

beorn7:
This release merely updates dependencies and, as a byproduct, the binaries that come with the release will be built with the current Go version (1.18.x).

The last release is quite long ago, and a lot of security issues have accumulated in depencies and the Go version used to build the binaries of the previous release. I do not now of any issues relevant to the Pushgateway, but at this point, it's probably better to cut a new release to avoid false alerts (and maybe even avoid an issue we don't know about yet). It comes with the risk of pulling in new bugs, though.


beorn7:
Sorry for weekend PR. I set the release date to Monday. Feel free to review when the weekend is over.

beorn7:
Fixes #469.

.circleci/config.yml
```diff
@@ -1,11 +1,12 @@
+---
 version: 2.1
 orbs:
   prometheus: prometheus/prometheus@0.16.0
 executors:
   # Whenever the Go version is updated here, .promu.yml should also be updated.
   golang:
     docker:
-      - image: circleci/golang:1.17
+      - image: quay.io/prometheus/golang-builder:1.18-base
```
SuperQ, line 9: We're actually switching these over to `cimg/go:1.18` except for `prometheus/prometheus`.

SuperQ (changes requested):
I would recommend changing the minimum Go version in go.mod to 1.17 and running a go mod tidy to get the new indirect dependency format.

beorn7:
Both comments addressed. Thank you very much. PHAL.

SuperQ (approved):
LGTM

roidelapluie (approved):

Issue #84, opened 2016-07-29
Feature request: Gzip support


y8:
Hi.

Right now pushgateway can't accept requests with `Content-Encoding: gzip`. 

It would be nice if `pushgateway` could to accept gzipped pushes, to reduce traffic.

brian-brazil:
If you've a use case where this matters, you may wish to reevaluate your use of the pushgateway.

See https://prometheus.io/docs/practices/pushing/

beorn7:
If somebody has a need for this even after having read and understood https://prometheus.io/docs/practices/pushing/ , I'd be willing to accept a PR. However, it is unlikely that a Prometheus core developer will take this on as none of us has the use case where it matters.

tompluess:
We have the same requirement here.
**Use case**: Several small devices should report their metrics to prometheus. these few devices are in heterogeneous networks, possible behind firewalls and may communicate to prometheus only through `https` as clients. Pushgateway seems to be the perfect solution for us. 
`Content-Encoding: gzip` would lower our network traffic drastically.

brian-brazil:
@tompluess  This is not an appropriate use of the pushgateway. You want something like pushprox.

beorn7:
Link to pushprox: https://github.com/RobustPerception/PushProx

mohnishkodnani:
If you have a batch job, where the collector Registry is in 100's of MB in the text format, would that be a usecase where we could consider gzip encoding.

beorn7:
The usual wisdom is that, if you really need to push 100s of MiB of metrics to the Pushgateway, you are not using it for its intended purpose.

But note that a PR that adds gzip support would be welcome, as said above. It's a feature that is of potential use for many use cases, including the intended ones, if network bandwidth is a scarce resource. I'm not going to block such a feature simply because it _also_ helps unintended use cases.

A more general note, however: In modern local networks, the CPU cycles to compress data are usually more expensive than the network bandwidth. https://github.com/kubernetes/kube-state-metrics, as an example, decided to disable gzip compression by default. It's not using the PGW, but it's offering humongous amounts of metrics for scraping. If the scrapes stays in the local network, it is both faster and less expensive if you disable compression.

belm0:
I'm finding that pushgateway performs poorly anyway when serving `/metrics` with large number of metrics.  (So for this issue, even if some bandwidth issue is addressed with gzip, you may then hit a scrape time bottleneck.)

[...]

PR #477, opened 2022-06-17
feat: add gzip support for /metrics

fatelei:
## What is pr do ?

Let pushgateway server handle gzip content-encoding content.

### Accept gzip

- server
image

- client
image

### Forbid gzip

- server
image

- client
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -390,6 +390,16 @@ guaranteed that the `DELETE` will be processed first (and vice versa).
 Deleting a grouping key without metrics is a no-op and will not result
 in an error.
 
+### Request compression
+
+The body of a POST or PUT request may be gzip-compressed. Add a header `Content-Encoding: gzip` to do so.
+
+Example:
+
+```
+echo "some_metric 3.14" | gzip | curl -H 'Content-Encoding: gzip' --data-binary @- http://pushgateway.example.org:9091/metrics/job/some_job
+```
+
 ## Admin API
 
 The Admin API provides administrative access to the Pushgateway, and must be
--- a/main.go
+++ b/main.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"compress/gzip"
 	"context"
 	"fmt"
 	"net"
@@ -56,6 +57,22 @@ type logFunc func(...interface{}) error
 
 func (lf logFunc) Println(v ...interface{}) {
 	lf("msg", fmt.Sprintln(v...))
+}
+
+func gunzipRequest(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		contentEncoding := r.Header.Get("Content-Encoding")
+		if contentEncoding == "gzip" {
+			gr, err := gzip.NewReader(r.Body)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+			defer gr.Close()
+			r.Body = gr
+		}
+		h.ServeHTTP(w, r)
+	})
 }
 
 func main() {
@@ -168,7 +185,7 @@ func main() {
 	})
 
 	mux := http.NewServeMux()
-	mux.Handle("/", r)
+	mux.Handle("/", gunzipRequest(r))
 
 	buildInfo := map[string]string{
 		"version":   version.Version,
--- a/api/v1/api.go
+++ b/api/v1/api.go
@@ -20,9 +20,8 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
-	"github.com/prometheus/common/route"
-
 	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/route"
 
 	"github.com/prometheus/pushgateway/handler"
 	"github.com/prometheus/pushgateway/storage"


main.go
```diff
@@ -16,6 +16,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/NYTimes/gziphandler"
```
beorn7, line 19: This import has to go into the block below (after line 32).

main.go
```diff
@@ -62,6 +63,7 @@ func main() {
 	var (
 		app                 = kingpin.New(filepath.Base(os.Args[0]), "The Pushgateway")
 		webConfig           = webflag.AddFlags(app)
+		gzipLevel           = app.Flag("gziplevel", "gzip compress level").Default("-1").Int() // gzip.DefaultCompression
```
beorn7, line 66: The doc string should be formatted as a sentence (start with capital letter, end with period), and maybe be a bit more descriptive (what levels are allowed, what's their meaning). See the other doc strings for comparison. Also, it seems to be more common to me to call it "compression level", not "compress level". And final nit: the flag itself should follow the structure of the other flags, so I would suggest "web.gzip-level" rather than "gziplevel".

main.go
```diff
@@ -109,12 +111,22 @@ func main() {
...
 	r.Get(*routePrefix+"/-/healthy", handler.Healthy(ms).ServeHTTP)
 	r.Get(*routePrefix+"/-/ready", handler.Ready(ms).ServeHTTP)
-	r.Get(
-		path.Join(*routePrefix, *metricsPath),
-		promhttp.HandlerFor(g, promhttp.HandlerOpts{
-			ErrorLog: logFunc(level.Error(logger).Log),
-		}).ServeHTTP,
-	)
+
+	gzHandler, _ := gziphandler.NewGzipLevelHandler(*gzipLevel)
```
beorn7, line 115: Users could provide an invalid level, so you do have to handle the error here.

main.go
beorn7: In different news, the default behavior should be to not do compression, even if the client requests it (because that's the current behavior, and compression might actually make things worse as discussed in #84, so we cannot just change behavior without cutting a major version).

I _think_ you can just pass in compression level 0 for "no compression", but I'm not sure if the gzip handler then really switches into "pass through" mode without overhead. So perhaps make the default here 0, and if the gzip handler handles that without overhead, we are good. Otherwise, you could simply add a condition below to not wrap with the gzip handler if the gzipLevel is 0.

beorn7 (changes requested):
Thanks for doing this. This seems to work in general, but most importantly, we have to add it in a way that the default behavior is the same as the current one (no compression, even if requested). See comments (which also include some nit-picking).

main.go
fatelei: ok

main.go
```diff
@@ -109,12 +111,22 @@ func main() {
...
+	metricHandler := promhttp.HandlerFor(g, promhttp.HandlerOpts{
+		ErrorLog: logFunc(level.Error(logger).Log),
+	})
+	if gzHandler != nil {
+		r.Get(
+			path.Join(*routePrefix, *metricsPath),
+			gzHandler(metricHandler).ServeHTTP,
```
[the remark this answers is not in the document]
fatelei: ok

beorn7 (changes requested):
Initially, you used github.com/NYTimes/gziphandler , but now you dropped it. Any reason for that? It seems reasonable to me to reuse that battle-tested middleware instead of implementing our own here.

In different news, you need to switch off gzip compression in the `promhttp.HandlerOpts` (line 137). Otherwise you might get double gzip compression.

main.go
```diff
@@ -58,10 +59,31 @@ func (lf logFunc) Println(v ...interface{}) {
...
+				} else {
+					r.Body = gr
+					h.ServeHTTP(w, r)
+				}
+			} else {
+				h.ServeHTTP(w, r)
+			}
+		})
+	}
+}
```
beorn7, line 80: Following thoughts:

- I think this doesn't have to be a doubly nested function. The outermost layer doesn't do anything.
- `ServeHTTP` won't call `Close` but someone has to close the gzip reader.
- I think we should only try to unzip if the Content-Encoding header is an exact match (who knows which future compression algorithms will contain "gzip").
- There is `http.Error` to write an error to a `ResponseWriter`.
- General style concern: Go tends towards avoiding the indented `else` branch if you can avoid it with a `return`.
- Finally, I'm thinking the name should better express that this is about request compression, not response compression. (Others might suffer the same confusion I did.) `gunzipRequest` perhaps?

In sum, this is my proposal:
```Go
func gunzipRequest(h http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		contentEncoding := r.Header.Get("Content-Encoding")
		if contentEncoding == "gzip" {
			gr, err := gzip.NewReader(r.Body)
			if err != nil {
				return
			}
			defer gr.Close()
			r.Body = gr
		}
		h.ServeHTTP(w, r)
	})
}
```

main.go
```diff
@@ -58,10 +59,31 @@ func (lf logFunc) Println(v ...interface{}) {
...
+
 func main() {
 	var (
 		app                 = kingpin.New(filepath.Base(os.Args[0]), "The Pushgateway")
 		webConfig           = webflag.AddFlags(app)
+		acceptGzip          = app.Flag("web.accept-gzip", "Web server can handle gzip content or not").Default("false").Bool() // gzip.DefaultCompression
```
beorn7, line 86: As mentioned above, I think we don't need a flag after all.

main.go
```diff
@@ -168,7 +190,11 @@ func main() {
 	})
 
 	mux := http.NewServeMux()
-	mux.Handle("/", r)
+	if *acceptGzip {
+		mux.Handle("/", decompressGzip()(r))
+	} else {
+		mux.Handle("/", r)
+	}
```
beorn7, line 197: So this can now be simply `mux.Handle("/", gunzipRequest(r))`.

beorn7 (changes requested):
My apologies for only realizing now that #84 only talks about gzip'ing the body of a _request_, which is not handled at all by the usual gzip handlers (which are only concerned about response compression).

I think I understand now. Good news is that it makes things easier in terms of behavior change. We don't even need a flag because the request encoding isn't really connected to content negotiations. So far, a gzip'd request was just failing, so we can simply always allow it now. The response encoding will be untouched by all of that.

For remaining things to do, see comments below that, hopefully, make more sense now. My apologies again.

Also, could you add a short section to the README.md about the new feature? Perhaps at the end of the [API section](https://github.com/prometheus/pushgateway#api)? Along the lines: "The body of a `POST` or `PUT` request may be gzip-compressed. Add a header `Content-Encoding: gzip` to do so." And then perhaps a curl example like `echo "some_metric 3.14" | gzip | curl -H 'Content-Encoding: gzip' --data-binary @- http://pushgateway.example.org:9091/metrics/job/some_job`

fatelei:
> My apologies for only realizing now that #84 only talks about gzip'ing the body of a _request_, which is not handled at all by the usual gzip handlers (which are only concerned about response compression).
> 
> I think I understand now. Good news is that it makes things easier in terms of behavior change. We don't even need a flag because the request encoding isn't really connected to content negotiations. So far, a gzip'd request was just failing, so we can simply always allow it now. The response encoding will be untouched by all of that.
> 
> For remaining things to do, see comments below that, hopefully, make more sense now. My apologies again.
> 
> Also, could you add a short section to the README.md about the new feature? Perhaps at the end of the [API section](https://github.com/prometheus/pushgateway#api)? Along the lines: "The body of a `POST` or `PUT` request may be gzip-compressed. Add a header `Content-Encoding: gzip` to do so." And then perhaps a curl example like `echo "some_metric 3.14" | gzip | curl -H 'Content-Encoding: gzip' --data-binary @- http://pushgateway.example.org:9091/metrics/job/some_job`

OK

main.go
fatelei: ok

main.go
fatelei: ok

main.go
fatelei: this code is more simple and clearly

README.md
```diff
@@ -390,6 +390,9 @@ guaranteed that the `DELETE` will be processed first (and vice versa).
 Deleting a grouping key without metrics is a no-op and will not result
 in an error.
 
+The body of a POST or PUT request may be gzip-compressed. Add a header Content-Encoding: gzip to do so.
```
beorn7, line 393: How about adding another header here? E.g. `### Request compression`. Otherwise, this appears to be part of the documentation for the `DELETE` method.

Please enclose `Content-Encoding: gzip` in back ticks.

README.md
```diff
@@ -390,6 +390,9 @@ guaranteed that the `DELETE` will be processed first (and vice versa).
...
+`echo "some_metric 3.14" | gzip | curl -H 'Content-Encoding: gzip' --data-binary @- http://pushgateway.example.org:9091/metrics/job/some_job`
```
beorn7, line 394: Please add a blank line before that. Otherwise it will all be in the same paragraph as the above. And perhaps start with "Example:". Then you could even use triple-backticks like this:

Example:
```bash
echo "some_metric 3.14" | gzip | curl -H 'Content-Encoding: gzip' --data-binary @- http://pushgateway.example.org:9091/metrics/job/some_job
```

beorn7 (approved):
Thank you very much again. Merging now.

PR #492, opened 2022-09-21
Update go.mod

mokafuri:
Signed-off-by: mokafuri <114009530+mokafuri@users.noreply.github.com>


beorn7:
Could you explain why this change is needed? The `go` directive has subtle effects, cf. https://go.dev/ref/mod#go-mod-file-go . Also, in this case, more changes would be needed to make the test pipeline pass.

PR #497, opened 2022-10-09
Misc code cleanup.

winder:
While inspecting the code I came across a few misc changes that seemed worth making.

### DIFF
--- a/api/v1/api.go
+++ b/api/v1/api.go
@@ -167,7 +167,7 @@ type response struct {
 
 func (api *API) respond(w http.ResponseWriter, data interface{}) {
 	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(200)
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 /pushgateway
 /.build
+/.idea
```
beorn7, line 3: We cannot really accommodate individual IDEs in .gitignore.

beorn7 (changes requested):
Thank you very much. All makes sense but let's not start to add IDE specific files to .gitignore (users of a specific IDE should put those files to their global gitignore).

.gitignore
winder: Got it, I have reverted this part of the PR.

beorn7 (approved):
Thanks again. Will merge on green…

PR #512, opened 2022-11-23
Cut v1.5.0

beorn7:
This also updates to exporter_toolkit v0.8.1, thereby introducing multiple listening addresses and systemd socket support. Please double-check that the code changes in `main.go` are valid.


SuperQ:
Want to copy `/.github/dependabot.yml` from the node_exporter repo? Would save some time keeping up with Go mod changes.

go.mod
```diff
@@ -3,35 +3,37 @@ module github.com/prometheus/pushgateway
...
-	google.golang.org/appengine v1.6.6 // indirect
-	google.golang.org/protobuf v1.26.0 // indirect
+	google.golang.org/appengine v1.6.7 // indirect
+	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-go 1.17
+go 1.19
```
SuperQ, line 39: Is there a reason to update this to 1.19? We typically try and support the last two releases.

main.go
```diff
@@ -211,13 +215,10 @@ func main() {
...
 
-	server := &http.Server{
-		Addr:    *listenAddress,
-		Handler: mux,
-	}
+	server := &http.Server{Handler: mux}
 
 	go shutdownServerOnQuit(server, quitCh, logger)
-	err = web.Serve(l, server, *webConfig, logger)
+	err := web.ServeMultiple(listeners, server, webConfig, logger)
```
SuperQ, line 221: Rather than use `ServeMultiple`, we could switch this to `web.ListenAndServe()`. Then we can drop the whole section at line 155.

go.mod
beorn7: Isn't that only a guideline for libraries but not for stand alone programs like the PGW?

Furthermore, setting this in `go.mod` doesn't prevent you from compiling with an older Go version. But it does change some other behavior. If I leave this at `go 1.17`, I get CI errors because `go mod tidy` has a different result than what I do locally. Maybe the CI actually uses Go 1.17 with the old directive for `go mod tidy`?

I don't understand the issue fully, but leaving this at `go 1.17` definitely creates an issue. If we can solve it in a different way, I'm all up for it.

beorn7: Update: I think I found a solution. Let's see what the CI says.

main.go
beorn7: I see. I changed it, and it seems to work.

SuperQ (approved):
LGTM

go.mod
SuperQ: Yea, there are some compat warnings with `go mod tidy`. I usually solve it by explicitly updating the thing it complains about.

beorn7:
\o/

PR #537, opened 2023-03-09
Add image build for s390x architecture

modassarrana89:
Updated Makefile for adding s390x architecture 
Updated .dockerignore  file for s390x

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@
 !.build/linux-armv7
 !.build/linux-arm64
 !.build/linux-ppc64le
+!.build/linux-s390x
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # limitations under the License.
 
 # Needs to be defined before including Makefile.common to auto-generate targets
-DOCKER_ARCHS ?= amd64 armv7 arm64 ppc64le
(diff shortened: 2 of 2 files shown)


beorn7:
I guess this should just work. It's similar to https://github.com/prometheus/prometheus/pull/6486 .

@simonpasquier you reviewed that PR back in the days. Do you see any concerns here?

I'll merge for now, but wait a bit before cutting a release.

beorn7 (approved):

beorn7:
For posteriority:

For the push to quay.io to work, I had to manually create the repository https://quay.io/repository/prometheus/pushgateway-linux-s390x

modassarrana89:
Pushgateway tar file is already s390x & same is being used in Quay . So this change was Just to build the image for pushgateway for s390x . This build is not for Quay

modassarrana89:
Thanks for the update . I can see s390x image for master release . Is there any plan to keep it as part of other release ex latest , etc

beorn7:
I will cut a new release in the next days, which will get an image than for an official release.

I do not plan to add images to older releases. There haven't been any invasive changes in the PGW for years, so I don't anticipate substantial need to run older versions instead of the current one.

PR #554, opened 2023-05-24
Cut v1.6.0

beorn7:


beorn7:
@SuperQ Note the new `verifyMetricFamily` helper function, which I think touches an issue you have run into, too.

handler/handler_test.go
```diff
@@ -619,3 +599,30 @@ func TestWipeMetricStore(t *testing.T) {
...
 	}
 }
+
+// verifyMetricFamily jumps through a few hoops because the current protobuf
+// implementation is deliberately creating an unstable formatting for the text
+// representation. So this takes the text representation of the expected
+// MetricFamily and unmarshals it into a proto message object first. Then it
+// marshals both the expected and the got proto message into a binary protobuf,
+// which it then compares.
+func verifyMetricFamily(t *testing.T, expText string, got *dto.MetricFamily) {
```
SuperQ, line 609: I wonder if we should add this helper somewhere in common.

SuperQ (approved):

SuperQ (commented):
I would also recommend copying `.github/workflows/golangci-lint.yml` from node_exporter. We stopped syncing this from `prometheus/prometheus`.

SuperQ (commented):
I would also recommend copying `.github/workflows/golangci-lint.yml` from node_exporter. We stopped syncing this from `prometheus/prometheus`.

go.mod
```diff
@@ -6,35 +6,34 @@ require (
 	github.com/golang/protobuf v1.5.3
```
SuperQ, line 6: This should be replaced with `google.golang.org/protobuf`

handler/handler_test.go
```diff
@@ -619,3 +599,30 @@ func TestWipeMetricStore(t *testing.T) {
...
+	gotProto, err := proto.Marshal(got)
+	if err != nil {
+		t.Errorf("unexpected error marshaling MetricFamily %v", got)
+	}
+
+	exp := &dto.MetricFamily{}
+	err = proto.UnmarshalText(expText, exp)
```
SuperQ, line 616: This can be updated to use `google.golang.org/protobuf/encoding/prototext`.

```suggestion
	err = prototext.Unmarshal([]byte(expText), exp)
```

SuperQ (changes requested):

handler/handler_test.go
beorn7: Yes, we should consider moving it up when it gets used there, too. However, I'm not quite sure yet if this is really the right way to do it… maybe some deep comparison without the final marshaling would be better?

SuperQ: The recommended way I saw was to use https://pkg.go.dev/github.com/google/go-cmp/cmp to compare with https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp.

beorn7: Oooof… two more dependencies instead of zero… I'd say we leave it as is in this PR and take some more time to consider our options for prometheus/common.

SuperQ: Yea, I'm not a fan of what this change requires.

roidelapluie (approved):

PR #564, opened 2023-07-12
Update link to the prom-aggregation-gateway

beorn7:
As you can see on
https://github.com/weaveworks/prom-aggregation-gateway , the maintenance has been moved to a fork.


SuperQ (approved):
Great, I'm glad someone has picked that up.

Issue #568, opened 2023-08-23
Add a reload button at the homepage or auto reload


thiago-marafeli-VBL:
## Feature request
When using the web interface for Pushgateway, whenever we want to reload the page, we have to reload the whole page (F5), but this makes all the dropdowns retract and back to the top of the page. Then, we have to search and open it all over again to see the value we want. It would be great if there was a button to reload the results without changing the "point of view" or an auto-reload one like the Grafana's, staying just as it is instead of scrolling all the way to the top of the page.

A convenient use case is when we are testing Pushgateway to watch the values and see if they are showing as expected at the page, before being introduced in Prometheus or Grafana.

ResamVi:
Hey I would take a crack at this for October - I'll look to add the reload button (auto-reload looks a bit advanced for now but could be a step for the next person)

beorn7:
The accordions are now re-opened upon a reload (initiated via the usual means, F5, or Ctrl-R or whatever you do). Thanks @ResamVi . An auto-reload with configurable timing as in Grafana would also be nice, but maybe not as important as for a dashboard. I'll close this for now, but whoever feels like adding an auto-reload options, please go for it.

PR #581, opened 2023-10-02
Keep record of unwinded containers on page refresh

ResamVi:
The solution retains a record of the currently un-collapsed container and unwinds it on a page refresh.
No extra refresh button and no live-update. 

The minor inconvenience of this solution is that on a page refresh the unwind animation plays.

[Screencast from 2023-10-02 18-40-31.webm](https://github.com/prometheus/pushgateway/assets/6261556/f0ac6de7-0364-425e-a947-81d4a737fcaa)

Helps #568

### DIFF
--- a/resources/static/functions.js
+++ b/resources/static/functions.js
@@ -119,11 +119,21 @@ $(function () {
 	    .removeClass('glyphicon-collapse-down')
 	    .addClass('glyphicon-collapse-up');
 	event.stopPropagation();
+
+	localStorage.setItem("coll_" + this.id, true);
     })
     $('div.collapse').on('hide.bs.collapse', function (event) {
 	$(this).prev().find('span.toggle-icon')
 	    .removeClass('glyphicon-collapse-up')
 	    .addClass('glyphicon-collapse-down');
 	event.stopPropagation();
+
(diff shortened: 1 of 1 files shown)


ResamVi:
@beorn7 would you have a look?

beorn7:
On my list.

beorn7:
I think this is quite good for a start.

But maybe we should use the session storage rather than the local storage? A user might open different windows with different views on the same PGW. Also, it might confusing if you come back to the PGW UI after some time and get elements expanded because they are still persisted in your local storage. You might also get garbage left behind, maybe even leading to multiple elements getting expanded on reload.

ResamVi:
You're right, I replaced it with `sessionStorage` now

beorn7 (changes requested):
Thank you. I think this will work as intended.

However, you need to run `make assets` and commit the results to bake the changes into the Go binary.

ResamVi:
> However, you need to run `make assets` and commit the results to bake the changes into the Go binary.

Alright, done. 

Thansk, I appreciate the guidance and your time in reviewing this 🙂

beorn7 (approved):
Thank you very much.

Issue #515, opened 2022-11-29
Add support for native histograms


beorn7:
It's still an experimental feature, but it should be relatively easy to add support to the PGW.

beorn7:
@jan--f is working on this.

jan--f:
So far I think this will just work as is. So things to be done are:

- [ ] add storage test for histograms
- [ ] add api test for histograms
- [ ] confirm that scraping in text format returns at least `_sum` and `_count`
- [ ] improve UI to display histograms

The last item might benefit from aligning with improvements made in Prometheus itself?

beorn7:
The Prometheus UI issue is https://github.com/prometheus/prometheus/issues/11269 . Currently, the UI just displays the `Histogram.String()` output. PGW can do the same for now. Note that the PGW UI uses its own bespoke query API to render the UI (`GET http://pushgateway.example.org:9091/api/v1/metrics` etc., see README.md). For now, it's probably fine to simply replace the `value` string, which is currently always a float in string form, with the string representation of the Histogram. For fancy graphical rendering, we needed to return the histogram in some structured form.

jan--f:
Displaying native histograms via `Histogram.String()` is in #596.

Otherwise pushgateway just works with native histograms.

Push via protobuf works as expected

![2023-11-22-150837_891x784](https://github.com/prometheus/pushgateway/assets/458501/8bead836-8681-4105-b535-3ca78a01d821)

Scraping via protobuf works as expected

![2023-11-22-150924_877x446](https://github.com/prometheus/pushgateway/assets/458501/0c249f02-032a-4418-9b1f-fa22e13cb68a)

Scraping with text format, give us an old-style histogram with a single bucket and a `_count` and `_sum` metric.

![2023-11-22-151429_852x485](https://github.com/prometheus/pushgateway/assets/458501/b9e1a61a-a559-4661-aa77-17ffaf07bb2c)

beorn7:
Thank you very much for all the work so far.

The representation in the PGW UI is the "machine level" one, not the human-readable one that you can see in the Prometheus UI (also see my comment on https://github.com/prometheus/pushgateway/pull/596).

[...]

PR #596, opened 2023-11-22
resources: for native histograms show String() output in the UI

jan--f:
If a histogram has no classic buckets, assume its a native histogram and render the stringyfied version of it.

Fixes: #515

### DIFF
--- a/handler/status.go
+++ b/handler/status.go
@@ -26,7 +26,9 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 
+	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/version"
+	"github.com/prometheus/pushgateway/histogram"
 	"github.com/prometheus/pushgateway/storage"
 )
 
@@ -72,6 +74,16 @@ func Status(
 				},
 				"base64": func(s string) string {
 					return base64.RawURLEncoding.EncodeToString([]byte(s))
+				},
+				"formatHistogram": func(m *dto.Histogram) string {
+					h, fh := histogram.NewModelHistogram(m)
+					if h == nil {
+						return fh.String()
+					}
+					return h.String()
+				},
+				"add": func(x, y int) int {
+					return x + y
 				},
 			})
 
--- a/histogram/prometheus_model.go
+++ b/histogram/prometheus_model.go
@@ -132,16 +132,16 @@ func GetAPIFloatBuckets(h *model.FloatHistogram) []APIBucket[float64] {
 }
 
 func makeBucket[BC model.BucketCount](bucket model.Bucket[BC]) APIBucket[BC] {
-	boundaries := uint64(2) // Exclusive on both sides AKA open interval.
+	boundaries := uint64(2) // () Exclusive on both sides AKA open interval.
 	if bucket.LowerInclusive {
 		if bucket.UpperInclusive {
-			boundaries = 3 // Inclusive on both sides AKA closed interval.
+			boundaries = 3 // [] Inclusive on both sides AKA closed interval.
 		} else {
-			boundaries = 1 // Inclusive only on lower end AKA right open.
+			boundaries = 1 // [) Inclusive only on lower end AKA right open.
 		}
 	} else {
 		if bucket.UpperInclusive {
-			boundaries = 0 // Inclusive only on upper end AKA left open.
+			boundaries = 0 // (] Inclusive only on upper end AKA left open.
 		}
 	}
 	return APIBucket[BC]{
--- a/resources/template.html
+++ b/resources/template.html
@@ -53,7 +53,6 @@ limitations under the License.
 			</div>
 		</div>
 	</nav>
-	
 	<div class="container-fluid" id="metrics-div">
 		{{- $data := .}}
 		{{- if eq (index .Flags "web.enable-admin-api") "true"}}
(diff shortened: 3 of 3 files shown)


resources/template.html
```diff
@@ -142,20 +142,26 @@ <h2 class="mb-0">
...
-				</tr>
-				{{- end}}
-				<tr>
-					<th scope="row">Total sample Count</th>
-					<td>{{.GetSampleCount}}</td>
-				</tr>
-				<tr>
-					<th scope="row">Sample Sum</th>
-					<td>{{value .GetSampleSum}}</td>
+					<td>{{.String}}</td>
```
beorn7, line 147: The problem here is that this is just the string representation of the protobuf, which is not very human-readable.

There should be code somewhere already to convert a protobuf histogram to `model.histogram.Histogram` or `model.histogram.FloatHistogram`. Maybe we can utilize it to get a display here that looks similar to what we currently get in the Prometheus UI itself?

I'm in a hurry right now, but I can help you to find this conversion code if you cannot find it easily.

jan--f: No worries, I'll go dig. Thanks for the pointer.

beorn7: So the code that does this during scraping is unfortunately too tightly coupled with the parser:
https://github.com/prometheus/prometheus/blob/980e2895a2eebb9664af104e6d4e19932cc74432/model/textparse/protobufparse.go#L173-L214

Maybe you could extract a separate `ProtoToHistogram(*dto.Histogram) (*histogram.Histogram, *histogram.FloatHistogram)` from that code (to be used there and then in the PGW, too).

beorn7: Or maybe not. In prometheus/prometheus, we use Gogo protobuf, in prometheus/pushgateway, we use the official Go protobuf library. So the dto.Histogram types are actually different.
I would just write the conversion code proto -> histogram.Histogram here in PGW for now. Depending on how long it is and where it is actually used, we can then move it somewhere else if needed.

beorn7:
Hi @jan--f , with #611 merged, it should just be one more small step to modify this PR to have a (more or less) human-readable histogram representation in the UI. Do you think you can work on this any time soon?

jan--f:
yes indeed, hoping to have an update next week.

jan--f:
Ok better late then never I suppose.
2024-04-19-114117_1920x802

Wdyt @beorn7

beorn7:
Looks great at first glance. I'm on vacation this week, but this is high on my list once I'm back in action.

handler/status.go
```diff
@@ -73,6 +75,20 @@ func Status(
 				"base64": func(s string) string {
 					return base64.RawURLEncoding.EncodeToString([]byte(s))
 				},
+				"formatHistogram": func(m *dto.Histogram) string {
+					h, fh := histogram.NewModelHistogram(m)
+					if h == nil {
+						return fmt.Sprintf("{ count:%v sum:%v %s}",
+							m.GetSampleCountFloat(),
+							m.GetSampleSum(),
+							histogram.BucketsAsString[float64](histogram.GetAPIFloatBuckets(fh)))
```
beorn7, line 84: Wouldn't `return h.String()` do the same?

handler/status.go
```diff
@@ -73,6 +75,20 @@ func Status(
...
+					} else {
+						return fmt.Sprintf("{ count:%v sum:%v %s}",
+							m.GetSampleCount(),
+							m.GetSampleSum(),
+							histogram.BucketsAsString[uint64](histogram.GetAPIBuckets(h)))
```
beorn7, line 89: And here `fh.String()`?

handler/status.go
```diff
@@ -73,6 +75,20 @@ func Status(
...
 					return base64.RawURLEncoding.EncodeToString([]byte(s))
 				},
+				"formatHistogram": func(m *dto.Histogram) string {
+					h, fh := histogram.NewModelHistogram(m)
+					if h == nil {
+						return fmt.Sprintf("{ count:%v sum:%v %s}",
+							m.GetSampleCountFloat(),
+							m.GetSampleSum(),
+							histogram.BucketsAsString[float64](histogram.GetAPIFloatBuckets(fh)))
+					} else {
```
beorn7, line 85: This `else` isn't really needed, is it? (You `return` in the `if` block above.)

histogram/prometheus_model.go
```diff
@@ -79,6 +80,32 @@ func BucketsAsJson[BC model.BucketCount](buckets []APIBucket[BC]) [][]interface{
 	return ret
 }
 
+func BucketsAsString[BC model.BucketCount](buckets []APIBucket[BC]) string {
```
beorn7, line 83: As said above, I assume you can use the `String` method of the model histograms, and then you don't need this helper.

resources/template.html
```diff
@@ -140,22 +139,29 @@ <h2 class="mb-0">
 				</tr>
 			</table>
 			{{- else}}
+                        {{- $h := .Histogram }}
 			{{- with .Histogram}}
 			<table class="table table-striped table-bordered">
-				{{- range .Bucket}}
+                                {{- if le (len .Bucket) 0}}
```
beorn7, line 145: So there is the case of a "dual histogram", which has both classic buckets and the sparse buckets of a native histogram, so that the scraper can decide what kind of histogram it wants to ingest (possibly even both). In that case, we should probably display both the classic and the native histogram, maybe with a `<th scope="row">Native Histogram representation</th>`.

beorn7 (changes requested):
Maybe I'm missing something, but I think this can be simplified, see comments.

go.mod
```diff
@@ -41,3 +41,5 @@ require (
 	google.golang.org/appengine v1.6.8 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
+
+replace github.com/prometheus/prometheus v0.48.1 => ../prometheus
```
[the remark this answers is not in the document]
jan--f: Indeed. Apologies... :facepalm:

handler/status.go
jan--f: That's the else branch if we don't have a histogram but a float histogram.

resources/template.html
jan--f: I see, so would then make sense to format classic histograms if there are classic buckets and native histograms if `Schema != nil`?

handler/status.go
jan--f: It sure does :+1:

resources/template.html
beorn7: Schema is just an int32, so it never can be nil.

I would render a classic histogram if there are classic buckets, and I would render a native histogram if there are native buckets. The only edge case is if there are no buckets at all, but then the rendering of either isn't much of a difference (just sum and count), so you could pick any one rendering for that.

handler/status.go
beorn7: I mean, don't remove the whole block, just remove the `} else {` and the closing  `}` below and outdent the content of the block.

resources/template.html
```diff
@@ -140,22 +139,31 @@ <h2 class="mb-0">
 				</tr>
 			</table>
 			{{- else}}
+                        {{- $h := .Histogram }}
 			{{- with .Histogram}}
 			<table class="table table-striped table-bordered">
-				{{- range .Bucket}}
+                                {{- if .Schema }}
```
beorn7, line 145: Hmm, I guess this  essentially tests if `.Schema` is 0 or not, which is not what we want. (Schema 0 is a perfectly valid schema for native histograms.) As said in another comment, let's simply test if there are native buckets.

resources/template.html
```diff
@@ -140,22 +139,31 @@ <h2 class="mb-0">
...
-					<td>{{.GetSampleCount}}</td>
-				</tr>
-				<tr>
-					<th scope="row">Sample Sum</th>
-					<td>{{value .GetSampleSum}}</td>
+                                    <th scope="row">Native Histogram representation</th>
+                                    <td>{{formatHistogram $h}}</td>
 				</tr>
+                                {{- end}}
+                                {{- if gt (len .Bucket) 0}}
```
beorn7, line 151: Or if there are neither native nor classic buckets (to handle the case of a bucket-less histogram).

beorn7:
Sorry for sprinkling the comments instead of doing them in one review. But I think I have addressed all questions now.

jan--f:
> Sorry for sprinkling the comments instead of doing them in one review. But I think I have addressed all questions now.

Not at all, thanks a lot for your patience. I think all comments are addressed now.

beorn7 (approved):
Wonderful. Thanks a lot.

Note that I'll update the commit description to the current state of affairs when merging.

PR #597, opened 2023-11-27
Add a flag for disabling pprof by default

cturiel:
Disabling "pprof" by default aligns with the principle of least privilege, reducing the attack surface without impacting functionality for users who do not require this endpoint.

@beorn7


beorn7:
> ... without impacting functionality for users who do not require this endpoint

But it _does_ impact the users that _do_ require this endpoint. :wink: 

It's a consistent practice across the Prometheus ecosystem to expose the pprof endpoint by default. We had discussed security impacts before, but concluded that most endpoints exposed by binaries from the Prometheus ecosystem are not meant to be accessible from untrusted networks anyway. pprof is often the least of a problem here.

Changing the default would be a breaking change that needed very serious security concerns to justify without a major version bump. Even adding a flag to disable the pprof endpoint if desired (but keep the default as it is now) wouldn't really make sense following the argument above.

See more discussions here: https://github.com/prometheus/pushgateway/discussions/520

As said there, this decision needed to be made across the whole project and should be discussed at the [dev summit](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit?pli=1). I'll close this for now, but we can re-open based on the outcome of the dev summit decision.

PR #616, opened 2024-01-18
Update dependencies and cut 1.7.0

beorn7:
The release will implicitly use Go1.21.6.


SuperQ (commented):
Can you remove `flags: -a -tags netgo` from `.promu.yml`? 

Also update `go 1.20` in `go.mod`.

SuperQ:
Not required, but the we determined that the flags weren't necessary anymore and the promu default is fine. Just cleanup.

The `go` update directive is needed by the new version of prometheus/common.

beorn7:
Thank you very much.

The prometheus/common dependency is updated to 0.46.0 even if I lease the `go` directive at 1.19.
As far as I understand, a recent Go compiler will compile the source code of each module according to its `go` directive. So we could also leave it at 1.19.

But since we don't get any breakage with 1.20 (I think Go will only introduce that for later releases), we can as well signal that you need a 1.20 capable compiler to compile at all because of prometheus/common.

beorn7:
Change pushed. PHAL.

SuperQ (approved):

PR #642, opened 2024-05-22
Push an abridged README to Docker Hub

beorn7:
The full README.md exceeds the character limit for the description on Docker Hub. This commit pushes an abridged README-docker.md to Docker Hub. It contains a reference to the full README.md.

### DIFF
--- a/README-containers.md
+++ b/README-containers.md
@@ -1,0 +1,78 @@
+# Prometheus Pushgateway
+
+[![CircleCI](https://circleci.com/gh/prometheus/pushgateway/tree/master.svg?style=shield)][circleci]
+[![Docker Repository on Quay](https://quay.io/repository/prometheus/pushgateway/status)][quay]
+[![Docker Pulls](https://img.shields.io/docker/pulls/prom/pushgateway.svg?maxAge=604800)][hub]
+
+The Prometheus Pushgateway exists to allow ephemeral and batch jobs to
+expose their metrics to Prometheus. Since these kinds of jobs may not
+exist long enough to be scraped, they can instead push their metrics
+to a Pushgateway. The Pushgateway then exposes these metrics to
+Prometheus.
+
+## Non-goals
+
+First of all, the Pushgateway is not capable of turning Prometheus into a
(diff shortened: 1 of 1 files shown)


.github/workflows/container_description.yml
```diff
@@ -13,7 +13,7 @@ permissions:
 jobs:
   PushDockerHubReadme:
     runs-on: ubuntu-latest
-    name: Push README to Docker Hub
+    name: Push an abridged README to Docker Hub
```
SuperQ, line 16: This file is synced from `prometheus/prometheus`, changes will be overwritten.

SuperQ:
Maybe we should move some of the text to other documentation files?

beorn7:
So far, having all the documentation for a relatively simple app just in README.md served us well.
Maybe it's time to move on to "real" documentation, but I don't want this to be motivated by a character limit on Docker Hub in combination with the inflexibility of our tooling.

beorn7:
Is there some way to implement an override for the GH actions? Should we create one?

beorn7:
Here is an idea: `readme_file: 'README.md'` is the default behavior anyway, so we could just leave it out. This has the additional twist that the tool will prefer a file called `README-containers.md` if it exists. (At least that's how I understand https://github.com/christian-korneck/docker-pushrm .) This behavior seems to be precisely tailored to the use case at hand here.

So plan would be:
1. Remove `readme_file: 'README.md'` from the source in `prometheus/prometheus`.
2. Let it propagate.
3. Rename `README-docker.md` in this PR to `README-containers.md` (and revert the other changes).

Only drawback is that it will deliver the shorter README also to Quay (which apparently allows larger descriptions), but I guess that's fine.

WDYT?

beorn7:
Chatted with @SuperQ , and the plan seems to work.

https://github.com/prometheus/prometheus/pull/14211 is merged for step 1. We have to wait until it has propagated (step 2), and then I will change this PR to implement step 3.

beorn7:
Updated according to the new plan.  @SuperQ PHAL.

PR #645, opened 2024-05-29
Log when liveness probes fails due to full write queue

vthemelis:


storage/diskmetricstore.go
```diff
@@ -125,7 +125,9 @@ func (dms *DiskMetricStore) Healthy() error {
 	// A pushgateway that cannot be written to should not be
 	// considered as healthy.
 	if len(dms.writeQueue) == cap(dms.writeQueue) {
-		return fmt.Errorf("write queue is full")
+		error_msg := "write queue is full"
+		level.Error(dms.logger).Log("msg", error_msg)
+		return fmt.Errorf(error_msg)
```
beorn7, line 130: In Go, local variables use camelCase. But I would just create the error directly. It will still log fine. Also, let's maybe use warn level, as you could legitimately be in an overloaded state and recover from it, and no harm was done.

In summary:

 ```suggestion
		err := fmt.Errorf("write queue is full")
		level.Error(dms.logger).Log("msg", err)
		return err
```

**Review by beorn7 — CHANGES_REQUESTED** [review]
Thank you for your contribution. Looks good in general, just some details to note.

**Comment by beorn7** [comment]
To get this into the next release, I'll apply the suggested changes directly here in the PR.

File: storage/diskmetricstore.go [hunk 2]
```diff
@@ -125,7 +125,9 @@ func (dms *DiskMetricStore) Healthy() error {
 	// A pushgateway that cannot be written to should not be
 	// considered as healthy.
 	if len(dms.writeQueue) == cap(dms.writeQueue) {
-		return fmt.Errorf("write queue is full")
+   	err := fmt.Errorf("write queue is full")
+   	level.Warn(dms.logger).Log("msg", err)
+   	return err
```
beorn7, line 130:
```suggestion
		err := fmt.Errorf("write queue is full")
		level.Warn(dms.logger).Log("msg", err)
		return err
```

beorn7 (approved):

PR #651, opened 2024-06-04
Print --version to stdout, not stderr

beorn7:
This get things in line with the Prometheus server.


aknuds1 (approved):

kunalmehta-eve:
@beorn7 Thanks for this, so we can expect new tag in next few days for the same ?

beorn7:
I'll cut a release this week or next.

PR #658, opened 2024-06-22
feat: add --web.disable-exporter-metrics flag

atompi:
If this flag is set, the metrics about the pushgateway itself (go_*, process_*, promhttp_*) will be excluded from /metrics.

This is useful in environments where you want to avoid recursive metric collection, i.e. you don't want the pushgateway's client_go default metrics to be exposed by itself.

Signed-off-by: atompi <atompi@foxmail.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -63,17 +63,18 @@ func (lf logFunc) Println(v ...interface{}) {
 
 func main() {
 	var (
-		app                 = kingpin.New(filepath.Base(os.Args[0]), "The Pushgateway").UsageWriter(os.Stdout)
-		webConfig           = webflag.AddFlags(app, ":9091")
-		metricsPath         = app.Flag("web.telemetry-path", "Path under which to expose metrics.").Default("/metrics").String()
-		externalURL         = app.Flag("web.external-url", "The URL under which the Pushgateway is externally reachable.").Default("").URL()
-		routePrefix         = app.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to the path of --web.external-url.").Default("").String()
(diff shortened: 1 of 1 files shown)


beorn7:
You should never remove those metrics from the /metrics endpoint. They are important to monitor the Pushgateway itself. 

If you use the PGW for what it is designed for, you will usually not run into naming collisions because the metrics exposed by the PGW itself are exactly those you would usually not push to a PGW.

If you still do so, you can still distinguish them by their labels. And if you really want to _not_ collect them, you can configure your Prometheus server to exclude them during the scrape.

Sorry to be blunt, but there will be no flag on the PGW to do something you should never do.

atompi:
@beorn7 I don't think those metrics are immovable. The node_exporter supports this parameter, and I think it's a good design.

beorn7:
The PGW is not an exporter.

atompi:
> The PGW is not an exporter.

@beorn7 So it should support this parameter as an application that optionally exposes its own metrics.

beorn7:
For one, it would require a flag of a different name, because it is not an exporter.

But the use case is also different. The Pushgateway is supposed to receive a handful of metrics, maybe a hundred, and they are meant to be precisely not the metrics that a binary usually exposes about itself (like process_, go_, promhttp_).

An exporter might export all kind of things, so _maybe_ the flag even makes sense (but only in weird edge cases). I don't want to burden the users of the PGW with the cognitive overhead of a flag that only serves a use case that is an antipattern for the PGW (namely blanket-push of many metrics te the PGW that are meant to be exposed directly).

PR #667, opened 2024-07-11
api: identify native histograms by presence if at least one span

jan--f:
We do the same already in the UI.


beorn7 (approved):
Actually, in the UI we are rendering both bucket versions if there are both. In the API, we only do native buckets or classic buckets.

I guess this is OK. We'll see if anyone complains. We have to keep this in mind if we should ever render the UI based on the API response.

PR #684, opened 2024-09-05
Update dependencies and Go

beorn7:
dependabot choked on the prometheus/prometheus update because of the `toolchain` directive in go.mod.
So I thought, let's update Go and all the dependencies properly.
I might cut a release with the now Go version soon, to pacify "security" scanners.
(Maybe I'll even do a bootstrap update for the same reason, but first I have to get this here done.)


beorn7:
Hmm, it looks CircleCI doesn't have go1.23 yet…

SuperQ:
No, we're still waiting on `cimg/go:1.23`. :crying_cat_face:

SuperQ:
Now that Go 1.23.1 is out, they should be able to fix their release.

beorn7:
We'll just wait a bit. I don't think this is very urgent.

beorn7:
Now it works! :tada:

SuperQ (approved):

Issue #623, opened 2024-03-05
Support UTF-8 in metric and label names


ywwg:
As raised in https://github.com/prometheus/client_ruby/issues/306, we will need some way to encode UTF-8 metric and label names in pushgateway URLs. This may be tricky because the current method does allow for encodings of job names and label *values*, and any solution would have to be backwards compatible with the current syntax. Thoughtful design work is needed.

See: https://github.com/prometheus/pushgateway#url

## Feature request
**Use case. Why is this important?**

[UTF-8 Support is coming to prometheus](https://github.com/prometheus/prometheus/issues/13095) and we need to cover this usecase.

fedetorres93:
The path for pushing metrics into the Pushgateway looks like

`/metrics/job/<JOB_NAME>{/<LABEL_NAME>/<LABEL_VALUE>}`

Currently, job names and label values can be encoded with base64url, in which case `job` or the label name must be suffixed with `@base64`. For example, using the grouping key `job="example",first_label="foo",second_label="bar"`:

```
/metrics/job@base64/ZXhhbXBsZQ==/first_label@base64/Zm9v/second_label@base64/YmFy
```

Some ideas I considered for encoding UTF-8 label names in URLs, using the same base64url approach:

- Encode the label name using base64url and add a new suffix (for example, `@base64name`) so that label names are decoded accordingly when the URL is parsed. We can have another suffix for the case where both the label name and the label value are base64url encoded, and keep the original `@base64` for when only the value is encoded. For example, using the grouping key `job="example",first.label="foo"`:
	
```
/metrics/job/example/Zmlyc3QubGFiZWw=@base64name/foo
```
	
- Use a query parameter (like `base64names=true`) to indicate label names are base64url encoded. This can also coexist with the current `@base64` suffix used for label values. Disadvantage: it will be all or nothing.

```
/metrics/job/example/Zmlyc3QubGFiZWw=/foo?base64names=true
```

Metric and label names in the request body are validated using functions from the `common` library that already support UTF-8.

ywwg:
is there content negotiation with calls to pushgateway, or do people just hit the endpoint?  (What if someone sends either of these formats to an older endpoint that doesn't understand the new grammar?)

fedetorres93:
> is there content negotiation with calls to pushgateway, or do people just hit the endpoint? (What if someone sends either of these formats to an older endpoint that doesn't understand the new grammar?)

[...]

PR #689, opened 2024-09-12
Add UTF-8 support in metric and label names

fedetorres93:
As part of the [initiative for adding UTF-8 support to Prometheus](https://github.com/prometheus/prometheus/issues/13095), this PR brings UTF-8 support for metric and label names as described in [this proposal](https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md) (for example: `{"metric.name", "label.name"="value"}`.

For UTF-8 label names defined in the URL path, [an escaping syntax](https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md#text-escaping) is used. For example, the label `"label.name"="value"` would be escaped like: `/metrics/job/example/U__label_2e_name/value`. This escaping is compatible with the base64 encoding for label values: `/metrics/job/example/U__label_2e_name@base64/dmFsdWU=`

UTF-8 support can be enabled by running the Pushgateway with the flag `--push.enable-utf8-names=true  `

Addresses https://github.com/prometheus/pushgateway/issues/623


beorn7:
Sorry, still fighting my backlog. I hope I get to review this tomorrow.

handler/handler_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/prometheus/common/model"
```
beorn7, line 21: This has to go into the next block (line 29).

main.go
```diff
@@ -17,6 +17,7 @@ import (
 	"compress/gzip"
 	"context"
 	"fmt"
+	"github.com/prometheus/common/model"
```
beorn7, line 20: As above, has to go to line 39 or so.

main.go
```diff
@@ -73,6 +74,7 @@ func main() {
 		persistenceFile     = app.Flag("persistence.file", "File to persist metrics. If empty, metrics are only kept in memory.").Default("").String()
 		persistenceInterval = app.Flag("persistence.interval", "The minimum interval at which to write out the persistence file.").Default("5m").Duration()
 		pushUnchecked       = app.Flag("push.disable-consistency-check", "Do not check consistency of pushed metrics. DANGEROUS.").Default("false").Bool()
+		pushEscaped         = app.Flag("push.enable-escaped-labels", "Allow escaped characters in label names in URLs.").Default("false").Bool()
```
beorn7, line 77: Let's call this `push.enable-utf8-names` to mirror the feature flag in Prometheus, and also to correctly describe what we are doing. (Without the flag, the PGW will reject non-legacy characters in names even in the payload of a push.)
Variable name and doc string has to be adjusted accordingly.

main.go
```diff
@@ -102,6 +104,10 @@ func main() {
 
 	ms := storage.NewDiskMetricStore(*persistenceFile, *persistenceInterval, prometheus.DefaultGatherer, logger)
 
+	if *pushEscaped {
+		model.NameValidationScheme = model.UTF8Validation
```
beorn7, line 108: Let's explicitly set `model.LegacyValidation` if the flag is not set (to defend against a change of the default in prometheus/client_model).

handler/push.go
```diff
@@ -180,19 +180,21 @@ func splitLabels(labels string) (map[string]string, error) {
 	for i := 0; i < len(components)-1; i += 2 {
 		name, value := components[i], components[i+1]
 		trimmedName := strings.TrimSuffix(name, Base64Suffix)
+		unescapedName := model.UnescapeName(trimmedName, model.ValueEncodingEscaping)
```
beorn7, line 183: We don't even want to try to unescape if the corresponding flag is not set. To not refer to the flag value directly here, maybe set a package-level variable of type `model.EscapingScheme` to either `model.ValueEncodingEscaping` or `model.NoEscaping` and then use that variable here.

handler/push.go
```diff
@@ -180,19 +180,21 @@ func splitLabels(labels string) (map[string]string, error) {
...
 		if !model.LabelNameRE.MatchString(trimmedName) ||
+			!model.LabelName(unescapedName).IsValid() ||
```
beorn7, line 185: With this validity test, wo don't need the one in the previous line anymore, do wo?

beorn7 (changes requested):
Thank you for doing this.
A few ideas about how to precisely control whether to escape or not. I hope I got it right, but happy to learn where I'm wrong.

In different news, this needs documentation (which currently is just in the ever growing README.md – feel free to add it there, breaking down the documentation into "proper" documentation files would be a task for another PR).

fedetorres93:
@beorn7 Thanks for your feedback, the requested changes were addressed.

handler/push.go
```diff
@@ -41,6 +41,8 @@ const (
 	Base64Suffix = "@base64"
 )
 
+var EscapingScheme = model.NoEscaping
```
beorn7, line 44: This could benefit from a doc comment.

beorn7 (changes requested):
Code looks good (modulu doc comment on `push.EscapingScheme`. But we also need documentation. (As suggested previously, I believe it is OK for now to add a section to the ever growing README.md, keeping in mind that eventually we would like to break it up into "proper" documentation.)

fedetorres93:
@beorn7 Sorry about the documentation, thanks for pointing it out, now it's added. Also added `EscapingScheme` comment.

README.md
```diff
@@ -320,6 +320,28 @@ Examples:
...
       /metrics/job/titan/name@base64/zqDPgc6_zrzOt864zrXPjc-C
 
+### UTF-8 support
+
+UTF-8 characters in metric and label names are supported by enabling
+the `--push.enable-utf8-names` flag. The syntax is as follows:
+
+    {"some.metric", "foo.bar"="baz"}
+
+For UTF-8 label names defined in the URL path, [an escaping syntax](https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md#text-escaping) is used.
```
beorn7, line 330: I would not explain the syntax of the text exposition format here. (It also doesn't cover protobuf.)

With some wordsmithing included, I would propose the following:

```suggestion
Newer versions of the Prometheus exposition formats (text and protobuf)
support the full UTF-8 character set in metric and label names. The
Pushgateway only accepts special characters in names if the command line
flag `--push.enable-utf8-names` is set.

To allow special characters in label names that are part of the URL path, the flag also enables a 
[specific encoding mechanism](https://github.com/prometheus/proposals/blob/main/proposals/2023-08-21-utf8.md#text-escaping). This is similar to the base64 encoding for label _values_ described above,
but works differently in detail for technical and historical reasons. As before, client libraries
should usually take care of the encoding. It works as follows:
```

README.md
```diff
@@ -320,6 +320,28 @@ Examples:
...
+
+* Prefix the string with `U__`.
+* All non-valid characters (i.e. characters other than letters, numbers, and underscores) will be encoded as underscores surrounding the Unicode value, like `_1F60A_`.
+* All pre-existing underscores will become doubled: `__`.
+* If a string should start with "U__" already, it will need to be escaped: `U___55_____`. (That's `U__` + `_55_` (for `U`) + `__` + `__`).
```
beorn7, line 335: Let's stay consistent with regard to the form. You start with an imperative form and then use future tense.

Here my suggestion with a few things added:

```suggestion
* A label name containing encoded characters starts with `U__`.
* All non-standard characters (i.e. characters other than letters, numbers, and underscores) are encoded as underscores surrounding their Unicode value, like `_1F60A_`.
* All pre-existing underscores are encoded as a double-underscore: `__`.
* If a label name starts with `U__` already, these characters have to be encoded as well, resulting in `U___55_____`. (That's `U__` + `_55_` (for `U`) + `__` + `__`).
* A label name starting with `U__` in it's encoded form, but containing invalid sequences (e.g. `U__in_xxx_valid`) is left unchanged.
```

README.md
```diff
@@ -320,6 +320,28 @@ Examples:
...
+
+For example, the label `"foo.bar"="baz"` would be escaped like:
```
beorn7, line 337: Let's better use "encoded" rather than "escaped" to only use one word throughout this file.

README.md
```diff
@@ -320,6 +320,28 @@ Examples:
...
+* Prefix the string with `U__`.
+* All non-valid characters (i.e. characters other than letters, numbers, and underscores) will be encoded as underscores surrounding the Unicode value, like `_1F60A_`.
+* All pre-existing underscores will become doubled: `__`.
+* If a string should start with "U__" already, it will need to be escaped: `U___55_____`. (That's `U__` + `_55_` (for `U`) + `__` + `__`).
+
+For example, the label `"foo.bar"="baz"` would be escaped like: 
+  
+    /metrics/job/example/U__foo_2e_bar/baz
+
+This escaping is compatible with the base64 encoding for label values:
```
beorn7, line 341: escaping → encoding, see above.

README.md
```diff
@@ -320,6 +320,28 @@ Examples:
...
+
+    /metrics/job/example/U__foo_2e_bar@base64/YmF6
+
```
beorn7, line 344: Let's add something about the one edge case, e.g.

Note that this method has an unlikely edge case that is not handled properly: A pusher unaware of the encoding mechanism might use a label name that is _also_ a valid encoded version of another label name. E.g. if a pusher intends to use the label name `U__foo_2e_bar`, but doesn't encode it as `U___55_____foo__2e__bar`, the Pushgateway will decode `U__foo_2e_bar` to `foo.bar`. This is the main reason why the decoding is opt-in via the `--push.enable-utf8-names` flag.

beorn7 (changes requested):
Cool. I just couldn't resist and applied a lot of wordsmithing. Please see my suggestions.

fedetorres93:
@beorn7 Thanks for the docs suggestions, done.

beorn7 (approved):
:ship:

PR #706, opened 2024-11-14
Update go.mod go security fixes

51n15t9r:
go 1.23.0 has some vulnerabilities that are fixed in 1.23.3 
Pushing that PR to address those.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/prometheus/pushgateway
 
 go 1.23
 
-toolchain go1.23.0
+toolchain go1.23.3
 
 require (
 	github.com/alecthomas/kingpin/v2 v2.4.0


beorn7:
There has been a lot of discussion around the `toolchain` directive, but I haven't noticed anyone proposing it should be used to prevent usage of Go versions with security issues.

The binaries released by the Prometheus projects are always built with the latest stable Go version (and I guess that means that we then auto-update the `toolchain` directive, as this is what the Go tools are doing these days – also signaling that we shouldn't update the directive manually).

It would be a huge change in our workflow to keep the `toolchain` directive up to date with the latest Go release that does not have a known security issue – and that would affect all of our many Go sub-projects, not just the Pushgateway. If you think we should do that, then I would like to ask you to propose it on the [prometheus-developers](https://groups.google.com/forum/#!forum/prometheus-developers) mailing list or on the #prometheus-dev channel on the CNCF Slack.

PR #735, opened 2025-03-19
Update Go and dependencies

beorn7:
Triggered by the failed dependabot PR https://github.com/prometheus/pushgateway/pull/734

Currently, I see no feasible way to avoid the usage of the deprecated global model.NameValidationScheme variable, as we cannot just default to UTF-8 support with PGW 1.x.


go.mod
```diff
@@ -1,16 +1,14 @@
 module github.com/prometheus/pushgateway
 
-go 1.23
-
-toolchain go1.23.0
+go 1.24
```
SuperQ, line 3:
```suggestion
go 1.23.0
```

beorn7: Hmm, I could have sworn this didn't work before, but now it does.

go.mod
```diff
@@ -1,16 +1,14 @@
...
-
-toolchain go1.23.0
+go 1.23.0
 
 require (
 	github.com/alecthomas/kingpin/v2 v2.4.0
 	github.com/golang/protobuf v1.5.4
-	github.com/golang/snappy v0.0.4
-	github.com/prometheus/client_golang v1.21.0
+	github.com/golang/snappy v1.0.0
```
SuperQ, line 8: Side note, it looks like this library is now archived.

SuperQ (approved):

go.mod
beorn7: Weird. They cut v1.0.0 just two weeks ago and then archived the repo, without any messaging around it in the Godoc or the Readme. Maybe the archival was accidental?

PR #740, opened 2025-04-02
Update README.md

mervecingoz:
windows powershell example is added

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Examples:
 
         curl -X PUT http://pushgateway.example.org:9091/api/v1/admin/wipe
 
-  #### MS Windows Users
+#### Note for MS Windows users
 
 MS Windows users can send HTTP requests to the Pushgateway using Powershell. 
 
--- a/README.md
+++ b/README.md
@@ -145,6 +145,15 @@ Examples:
 
         curl -X PUT http://pushgateway.example.org:9091/api/v1/admin/wipe
 
+  #### MS Windows Users
+MS Windows users could send http requests to the pushgateway using Powershell. 
+
+Example:
+
+         Invoke-WebRequest -Uri http://localhost:9091/metrics/job/MyJob -Method POST -Body "Some_Metric 3.14`n"
+
+Double quotes should be used to enclose the metric text, and the back tick (`) character should come before the line ending character (n) at the end.
+
 ### About the job and instance labels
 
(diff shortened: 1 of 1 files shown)


beorn7:
From the KubeCon Contrib Fest. @bwplotka, @ArthurSens and @vesari , I'll take this one.

beorn7:
Sorry for the delay. I was busy during KubeCon. I'll get to this ASAP.

README.md
```diff
@@ -145,6 +145,15 @@ Examples:
 
         curl -X PUT http://pushgateway.example.org:9091/api/v1/admin/wipe
 
+  #### MS Windows Users
```
beorn7, line 148: Let's not indent this. I think it looks better without it. Also, we are not really using title case in the sub-titles. And maybe add something like "Note for". In summary:

```suggestion
#### Note for MS Windows users
```

README.md
```diff
@@ -145,6 +145,15 @@ Examples:
...
+MS Windows users could send http requests to the pushgateway using Powershell.
```
beorn7, line 149: I would leave a blank line above this for consistency. Similarly, capitalize "Pushgateway" and "HTTP".

beorn7 (changes requested):
Thank you very much. Content looks good. I just have a few nits about formatting and such.

README.md
beorn7: Final nit: Maybe "can" instead of "could"?

beorn7:
@mervecingoz there is very little left to do here. Do you think you'll come back to it?

fpastorh:
@beorn7 I can do those small modifications, I was working on this as well on the contribfest but I don't think I have the rights to change anything on this PR. If you approve this PR I can create a new one with the suggested changes.

beorn7:
Thanks for your offer @fpastorh . I don't think we should jump through that many hoops for two simple changes. If @mervecingoz doesn't come back to this soon, I'll just submit the changes via the suggestion mechanism here in the PR.

README.md
beorn7:
```suggestion

MS Windows users can send HTTP requests to the Pushgateway using Powershell. 
```

beorn7 (approved):

beorn7:
Thinking of it, that's easy enough. I have just done that.

fpastorh:
Looks good!

PR #811, opened 2026-03-31
Upgrade Bootstrap to v5

### LINKED ISSUE #687: Update Bootstrap to v5
While the know security problems of the used Bootstrap library (4.3.1) do not affect the Pushgateway, it would be good to update to Bootstrap v5 eventually.

As far as I can see, v5 has dropped the Glyphicons support, so an update is definitely more involved than just downloading Bootstrap 5 and put it into the resources directory.

dswarbrick:
Replace deprecated Bootstrap v4 library with v5.

Closes: #687


beorn7:
I don't have strong feelings here.

But this should be reviewed/decided by the new maintainers @jan--f and @machine424 .

resources/static/functions.js
jan--f, line 66: Does this need the same change as lines 34 and 42?

jan--f, line 86: Same here?

jan--f (commented):
Generally this looks pretty good. I have one question below.

I think both you suggestions (the glyphicon and jquery dependecy) would make good follow ups. Would you perhaps even be interested in adding those?

PR #819, opened 2026-04-30
Update dependencies to fix security issues

bartoszkosiorek:


machine424:
Thanks for the PR,
Dependabot is set up to take care of updates (I've just merged some), could you elaborate on why we need to do this manually?

junoatwork:
> Dependabot is set up to take care of updates (I've just merged some)

Thanks for your help and attention keeping the repo up-to-date!

Found this while checking if there is an updated release. While the master branch indeed has clean dependencies, the gh releases are in a bit worse shape.

Looking at the latest  v1.11.2 -  The highest severity is `crypto/tls`, but several of the builtin go packages the release is built with have active CVEs. Additionally the github.com/prometheus/prometheus version v0.307.3 in that release is out of date and has several vulnerabilities.

For now, my workaround is to take on building from source.

Is there any interest in further automating the release process to stay up-to-date? The two big gaps are automating the build trigger and updating the go version in addition to the package dependencies.

Again, thank you!

ywwg:
Can you also bump the go version to match prometheus, go version 1.25.6?

ywwg:
ah it looks like that prom release used 1.25.0 so this is correct.

ywwg (approved):
lgtm

ywwg (changes requested):
discussing with @roidelapluie , retracting lgtm because this may be redundant

roidelapluie:
Thanks! 1.11.3 was released with the changes.

PR #824, opened 2026-05-13
ci: pin contents: read on the CI workflow

arpitjain099:
Pins the default `GITHUB_TOKEN` to read-only for `ci.yml`. Image push (Docker Hub / Quay) and release publication use their own dedicated secrets (`docker_hub_login`/`docker_hub_password`, `quay_io_login`/`quay_io_password`, and the prom-bot PAT for releases), so the default token only needs read access for the checkout. Matches the top-level pattern in the other prometheus repos.

PR #826, opened 2026-05-25
ci: cap `GITHUB_TOKEN` to `contents: read`

arpitjain099:
The `ci` workflow runs Go build / test only. No GitHub API writes, so workflow-level `contents: read` is the right cap for the default `GITHUB_TOKEN`.

Post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
     branches: [master, 'release-*']
     tags: ['v*']
 
+permissions:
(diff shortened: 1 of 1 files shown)


jan--f:
Seems to be the same as #824

arpitjain099:
Yes you're right @jan--f, same as #824 (which already merged). Closed this one myself; apologies for the noise.

PR #866, opened 2026-09-14
Bump prometheus/common to v0.71.0 to stop formatting discarded logs

### LINKED ISSUE #863: performance: High latency increase after upgrading from 1.10 to 1.11
Hi after upgrade from 1.10 to 1.11 we were seeing huge latencies on pgw. 

Unfortunately some of our pushgateways have up to 15k series and around 3 pushes per second. 
_(I know that's not the intended use-case… I'm not happy with it, but developers want it this way)_

Anyway, the bigger the instance the bigger the latency impact was. 
After investigation, it turned out the root cause is logging.
We bumped into this earlier when info level logs about inconsistent HELP of metrics caused the pgw to eat up lots of CPU so we increased log level to warn to mitigate this.

BUT after upgrading to 1.11 it turns out switch to `promslog` lead to a regression, that the CPU is wasted even if the log level is `warn`.

Pprof with `--log-level=warn` 👇 See the `log/slog.(*Logger).Info` even on `warn` log level

```go tool pprof -top -cum cpu.prof | head -20
File: pushgateway
Build ID: c1b7e44f1f1913a38626129896731b8cadd7e49e
Type: cpu
Time: 2026-08-13 18:16:55 CEST
Duration: 120s, Total samples = 16.89s (14.07%)
Showing nodes accounting for 13.49s, 79.87% of 16.89s total
Dropped 353 nodes (cum <= 0.08s)
      flat  flat%   sum%        cum   cum%
         0     0%     0%     11.06s 65.48%  github.com/prometheus/client_golang/prometheus.Gatherers.Gather
         0     0%     0%     10.90s 64.54%  github.com/prometheus/pushgateway/storage.(*DiskMetricStore).checkWriteRequest
         0     0%     0%     10.90s 64.54%  github.com/prometheus/pushgateway/storage.(*DiskMetricStore).loop
         0     0%     0%     10.77s 63.77%  github.com/prometheus/client_golang/prometheus.GathererFunc.Gather
         0     0%     0%     10.77s 63.77%  github.com/prometheus/pushgateway/storage.(*DiskMetricStore).GetMetricFamilies
         0     0%     0%     10.75s 63.65%  github.com/prometheus/pushgateway/storage.(*DiskMetricStore).checkWriteRequest.func2
         0     0%     0%     10.72s 63.47%  log/slog.(*Logger).Info (inline)
         0     0%     0%     10.72s 63.47%  log/slog.(*Logger).log
         0     0%     0%     10.72s 63.47%  log/slog.(*TextHandler).Handle
         0     0%     0%     10.72s 63.47%  log/slog.(*commonHandler).handle
         0     0%     0%     10.70s 63.35%  log/slog.(*handleState).appendAttr
         0     0%     0%     10.70s 63.35%  log/slog.(*handleState).appendValue
         0     0%     0%     10.70s 63.35%  log/slog.appendTextValue
         0     0%     0%     10.69s 63.29%  log/slog.(*handleState).appendNonBuiltIns
         0     0%     0%     10.69s 63.29%  log/slog.(*handleState).appendNonBuiltIns.func1 (inline)
         0     0%     0%     10.69s 63.29%  log/slog.Record.Attrs (inline)
         0     0%     0%      8.15s 48.25%  fmt.Sprintf
```

After digging deeper it turned out that the `promslog.NopLogger` actually does serialize the log message and after that discards it, so solution would be dropped it right away. 

The main pain point is its usage in the `DiskMetricStore` and logging the inconsistent help strings

https://github.com/prometheus/pushgateway/blob/bc06b7e79aba7724055ba40631a90f8f84d16936/storage/diskmetricstore.go#L381

https://github.com/prometheus/pushgateway/blob/bc06b7e79aba7724055ba40631a90f8f84d16936/storage/diskmetricstore.go#L167

I created a PR in the prometheus/common with this fix https://github.com/prometheus/common/pull/960

if accepted all that would be needed on the pgw side is to update the dependency, creating the issue more like a tracker/reminder

KR-Ravindra:
## Problem

Since Pushgateway 1.11 switched to `promslog`, push latency grows with the size of the store as soon as any stored metric family has an inconsistent help string, even with `--log.level=warn`. The reporter's profile in #863 shows ~63% of CPU in `slog.(*TextHandler).Handle` under `checkWriteRequest`.

## Root cause

`storage/diskmetricstore.go:381`: `checkWriteRequest` runs the consistency check on a scratch `DiskMetricStore` with `promslog.NewNopLogger()`. Up to `prometheus/common` v0.70.x that logger is a `TextHandler` writing to `io.Discard` whose `Enabled()` returns true, so the `Info` calls at `storage/diskmetricstore.go:167` and `:175`, which pass whole `*dto.MetricFamily` values as attributes, serialize every metric of the affected families on every push and then throw the text away.

## Fix

Bump `github.com/prometheus/common` to v0.71.0 (`go get` + `go mod tidy`). prometheus/common#960, released in v0.71.0, makes `NewNopLogger` return `slog.New(slog.DiscardHandler)`, which is disabled at every level, so the attributes are never formatted. Dependabot's #861 only proposes v0.70.1, which predates that fix.

Adds `TestCheckWriteRequestInconsistentHelpAllocs`, which pushes a single metric against a stored 1000-metric family and asserts that an inconsistent help string does not cost more allocations than a consistent one.

## How tested

Before (common v0.69.0):

```
--- FAIL: TestCheckWriteRequestInconsistentHelpAllocs (0.10s)
    diskmetricstore_test.go:1615: checkWriteRequest with inconsistent help strings allocates 34115 more than with consistent help strings (39130 vs 5015), metric families are formatted although logging is disabled
```

After (common v0.71.0): the test passes (5009 vs 5009 allocations), `go test ./...`, `go vet ./...`, `go build ./...` and `make common-lint` (golangci-lint v2.12.2) are clean.

## Links

Fixes https://github.com/prometheus/pushgateway/issues/863
Upstream fix: https://github.com/prometheus/common/pull/960

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.
1. prometheus/common v0.71.0 ships #960 (`promslog.NewNopLogger` now returns `slog.New(slog.DiscardHandler)`); v0.70.1 still formats through a TextHandler into `io.Discard`, so dependabot's #861 alone would not fix #863.
2. `go mod verify` and `go mod tidy` are clean on this branch; the other go.mod/go.sum changes are transitive requirements of common v0.71.0.
3. With common pinned back to v0.69.0 and v0.70.1 the new test fails at `diskmetricstore_test.go:1615` with ~34k extra allocations; it passes on v0.71.0. `go test ./...`, `go vet`, `gofmt` clean locally.
4. This conflicts with #861 on go.mod/go.sum; I will rebase whichever lands second.
Marking ready for review.

PR #869, opened 2026-09-17
Bump golang.org/x/net and golang.org/x/crypto to fix known CVEs

### LINKED ISSUE #868: Update dependencies to fix CVE-2026-46600 and CVE-2026-56854
[Scan Info]
```
Vulnerable Component: go://golang.org/x/net:0.55.0
Fix Versions: 0.56.0
```

[Scan Info]
```
Vulnerable Component: go://golang.org/x/crypto:0.52.0
Fix Versions: 0.55.0
```

pujitha24:
Motivation:
A vulnerability scan flagged two indirect dependencies pulled in via
github.com/prometheus/exporter-toolkit/web and
github.com/prometheus/common/config:
  - golang.org/x/net v0.55.0 (CVE-2026-46600, fixed in v0.56.0)
  - golang.org/x/crypto v0.52.0 (CVE-2026-56854, fixed in v0.55.0)

Approach:
Ran `go get golang.org/x/net@v0.57.0 golang.org/x/crypto@v0.55.0`
followed by `go mod tidy`. x/net had to go to v0.57.0 rather than the
minimum v0.56.0 because golang.org/x/crypto@v0.55.0's own go.mod
requires golang.org/x/net@v0.57.0 (confirmed via `go mod graph`); this
is Go's minimal version selection, not a manual choice. `go mod tidy`
also bumped the transitively-required golang.org/x/sync, golang.org/x/sys,
and golang.org/x/text to versions compatible with the new x/net/x/crypto.
No source files were touched; both packages are only reachable
transitively, pushgateway's own code never imports them directly.

A newer golang.org/x/crypto v0.56.0 is available and additionally fixes
three golang.org/x/crypto/ssh advisories, but bumping to it also raises
this module's go.mod `go` directive from 1.25.0 to 1.26.0, which is a
larger, unrelated change. pushgateway does not import the ssh package
(govulncheck confirms 0 reachable vulnerabilities either way), so this
change stays at v0.55.0 to keep the fix minimal and scoped to the two
reported CVEs.

Validation:
- `go build ./...` succeeds.
- `go test ./...` passes (api/v1, handler, storage).
- `go mod tidy` produces no further diff (go.mod/go.sum are tidy).
- `golangci-lint run` reports 0 issues (run with a local golangci-lint
  v2.13.2 build; the repo's own `make lint` target auto-downloads the
  pinned v2.12.2 binary, which is built with go1.26 and panics against
  this machine's go1.27.1 toolchain - a local environment mismatch,
  not something this change causes. Upstream CI runs inside the
  quay.io/prometheus/golang-builder:1.26-base container, so this
  mismatch will not occur there).
- `govulncheck` reports "Your code is affected by 0 vulnerabilities"
  both before and after this change, since neither package is called
  directly; the bump removes the flagged modules from the dependency
  list regardless.

Report: https://github.com/prometheus/pushgateway/issues/868
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #868


rogov-sa (approved):
Lgtm

